PR #28, opened 2015-01-25
Add an option to instruct the RL-glue interface to send the RGB components for each pixel

alito:
The RL-glue interface currently sends the screen as a series of values between 0 and 127 which correspond to the internal pallette used by the game. While this is useful for distinguishing colours between each other, it doesn't let the client-side see the actual colours of the screen.

This patch adds an option (-send_rgb), which defaults to false, that switches the output sent through the RL-glue interface to the RGB values of the pixels, each component consisting of a value (0-255). The output then becomes: red value for pixel 1, green value for pixel 1, blue value for pixel 1, red value for pixel 2, etc.

Since this defaults to false, the behaviour of ALE doesn't change unless extra parameters are given to it.

The documentation (Settings.cxx and manual.tex) was modified to reflect this added option.


src/controllers/rlglue_controller.cpp
```diff
@@ -244,8 +256,22 @@ reward_observation_terminal_t RLGlueController::constructRewardObservationTermin
   // Copy RAM and screen into our big int-vector observation 
   for (size_t i = 0; i < ram.size(); i++)
     m_observation.intArray[index++] = ram.get(i);
-  for (size_t i = 0; i < screen.arraySize(); i++)
-    m_observation.intArray[index++] = screen.getArray()[i];
+
+  if (m_osystem->settings().getBool("send_rgb")) {
```
mgbellemare: There's a possible buffer overrun here if the send_rgb flag is modified after creating the intArray. One solution is to store send_rgb as a const member variable and rely on this; or at least assert that the intArray size matches arraySize * 3.

doc/manual/manual.tex
```diff
@@ -176,6 +177,17 @@ \subsection{Internal Interface Arguments}
 \end{verbatim}
 }
 
+\subsection{RL-Glue Interface Arguments}
+
+\small{
+\begin{verbatim}
+  -send_rgb [true|false] -- if true, RGB values are sent for each pixel
```
mgbellemare: The rest of the manual uses < > for arguments (I do realize the -help printout is inconsistent with this! probably this needs to be fixed). Still, we should use < > here.

alito:
Good points. I've updated the manual and added an assert. I can go the member variable route if you'd rather.

mgbellemare:
Great, thanks. I realize now we should possibly throw an exception here, since some compilers remove asserts (e.g. Visual Studio compiling in release mode). But we can do proper exception handling when the code is cleaned up (which should happen over the next few months).

PR #32, opened 2015-02-10
Refactored RomSettings to remove the ActionVects.

mhauskn:
This PR corrects an issue in which ALE would not update the set of legal & minimal actions if a second rom was loaded.  

Example of Problem (using C++ Interface): 
```c++ 
    ale.loadROM("skiing.bin");
    ActionVect legal_actions = ale.getLegalActionSet();
    for (int i=0; i<legal_actions.size(); ++i) {
      cout << legal_actions[i] << ", ";
    }
    cout << endl;

    ale.loadROM("space_invaders.bin");
    legal_actions = ale.getLegalActionSet();
    for (int i=0; i<legal_actions.size(); ++i) {
      cout << legal_actions[i] << ", ";
    }
    cout << endl;
```
Correct Output:
```
Game console created:
  ROM file:  /u/mhauskn/projects/ale-assets/roms/skiing.bin
...
0, 2, 3, 4, 5, 6, 7, 8, 9, 

Game console created:
  ROM file:  /u/mhauskn/projects/ale-assets/roms/space_invaders.bin
...
0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17
```

Current Output:
```
Game console created:
  ROM file:  /u/mhauskn/projects/ale-assets/roms/skiing.bin
...
0, 2, 3, 4, 5, 6, 7, 8, 9, 

Game console created:
  ROM file:  /u/mhauskn/projects/ale-assets/roms/space_invaders.bin
...
0, 2, 3, 4, 5, 6, 7, 8, 9,
```
Explanation: RomSettings stored ActionVects for legal and minimal actions which were filled in correctly when empty but not changed if unempty. In the case above, when a new rom is loaded, these members retain their old values and are never updated, causing them to return stale values on subsequent function calls.

Proposed solution: I've removed the ActionVect class members and instead just create + return an action vect when the get_legal_actions/get_minimal_actions methods are called. Since no state is preserved RomSettings  returns the correct ActionVects for whatever Rom is currently being played. Creating a new ActionVect may be a bit slower, but this call is typically only done once per rom during setup so the additional overhead is likely negligible. 

Proposed solution was tested against the internal agent and the c++ shared library agent.

### DIFF
--- a/src/agents/PlayerAgent.hpp
+++ b/src/agents/PlayerAgent.hpp
@@ -77,7 +77,7 @@ protected:
     int episode_frame_number;
     int episode_number;
         
-    ActionVect & available_actions;
+    ActionVect available_actions;
 
     bool record_trajectory;
     ActionVect trajectory;
--- a/src/games/RomSettings.cpp
+++ b/src/games/RomSettings.cpp
@@ -17,32 +17,28 @@
  */
 #include "RomSettings.hpp"
 
-ActionVect RomSettings::actions;
-ActionVect RomSettings::all_actions;
-
 bool RomSettings::isLegal(const Action& a) const {
   return true;
 }
 
-ActionVect& RomSettings::getMinimalActionSet() {
-  if (actions.empty()) {
-    for (int a = 0; a < PLAYER_B_NOOP; a++)
-      if (isMinimal((Action)a) && isLegal((Action)a))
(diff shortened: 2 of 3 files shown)


mgbellemare:
Yes, that's a far better solution. LGTM.

PR #33, opened 2015-02-18
Added legacy constructor to ale_interface.

mhauskn:
Tested the legacy constructor in the shared_library_interface_example and confirmed that it works as expected.


mgbellemare:
LGTM.

PR #34, opened 2015-02-19
Added CMake support!

mhauskn:
Tested on Ubuntu and OSX.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,60 @@
+cmake_minimum_required (VERSION 2.6)
+project(ale)
+
+option(USE_SDL "Use SDL" OFF)
+option(USE_RLGLUE "Use RL-Glue" OFF)
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+cmake_minimum_required (VERSION 2.6)
+project(ale)
+
+option(USE_SDL "Use SDL" OFF)
+option(USE_RLGLUE "Use RL-Glue" OFF)
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-multichar -Wunused -fno-rtti -fPIC -O3 -fomit-frame-pointer")
```
mgbellemare: Do we need no-rtti? The internet suggests this is only useful to save a bit of space, and it kills dynamic_cast, which on the long run could cause someone headaches.

Also (and somewhat pedantically), why -Wno-multichar?

mhauskn: These were just the flags used by the unix makefile. Please remove if you
want to

PR #37, opened 2015-02-20
Much needed display_screen update

mhauskn:


mhauskn:
Changes from above PR are:
1) Now only requires libSDL - no more sdl_gfx/sdl_image/sdl_rotozoom etc.
2) Fixed the resolution to 192x160 which I hear is standard for Atari 2600. Err actually 384x320 so that the screen is not too small. This give the screen a much more realistic looking appearance.
3) Built manual keyboard control into display screen. With the display window up, just press 'm' to enter manual control mode :)
4) Cleaned up the interfaces and removed legacy garbage linked to multiple SDL_handlers.

mgbellemare:
Mmm. The (core) resolution should be 192 rows (== scanlines) by 160 columns. See e.g. https://atariage.com/2600/programming/2600_101/02breach.html .

What I've done typically is double the width, and then scale by 2 (so in your case this would be 192 x 640). I suppose if you wanted to emulate a 4:3 TV then this isn't quite right, but it's easier to handle.

mhauskn:
Cool, can you take a look at this once you've got SDL working? 

I had been ballparking the resolution to match what I see on youtube videos of people playing on the actual Atari 2600. This video (http://www.youtube.com/watch?v=NqyKpJwk29o) has space invaders alien sprites that are nearly square. The 384x320 resolution I have now mimics this pretty well.

mgbellemare:
Ahh, it does look pretty. Have you tried 400x300 (or, closer to what you previously had: 428x321)? This matching the TV's 4:3 aspect ratio. Here's a screenshot from Stella: it is in 4:3.
spaceinvaders

mhauskn:
I just checked out 428x321 and think it looks good. Can you make the change
and push?

On Sat, Feb 28, 2015 at 4:53 PM, Marc G. Bellemare <notifications@github.com

mgbellemare:
Done.

PR #40, opened 2015-02-22
Added record_screen_dir argument. Refactored ALEInterface set methods.

mhauskn:
This PR does 3 things:

1) Cleaned up and standardize the -help printout for ALE. It now all fits on one screen :)
2) Added -record_screen_dir argument that uses StellaEnvironment to automatically save each screen as a png to the specified directory. This should make the process of making ALE movies alot easier. 
3) Changed the set* methods in ALEInterface to specify what type of data they are handling (e.g. new methods are setInt/setBool/setString/setFloat). This was necessary because c++ was incorrectly inferring the type of data I was passing. In sharedLibraryInterface.cpp I would call "ale.set("record_screen_dir", "screens");" and it would infer that "screens" was a bool rather than a string. Adding the (Int/Bool/String/Float) suffixes to the set functions removes the possibility of incorrect type inference.


src/environment/stella_environment.cpp
```diff
@@ -130,6 +136,14 @@ reward_t StellaEnvironment::act(Action player_a_action, Action player_b_action)
     sum_rewards += oneStepAct(player_a_action, player_b_action);
   }
 
+  if (!m_record_screen_dir.empty()) {
+    static int recorded_frame = 0;
+    std::stringstream ss;
+    ss << m_record_screen_dir << "/" << setfill('0') << setw(6) <<
+        std::to_string(recorded_frame++) << ".png";
```
mgbellemare: Is there a reason to use std::to_string rather than pass recorded_frame to the stringstream? std::to_string is new in C++11.

src/ale_interface.cpp
```diff
@@ -147,25 +147,25 @@ float ALEInterface::getFloat(const std::string& key) {
...
 // Set the value of a setting.
-void ALEInterface::set(const string& key, const string& value) {
+void ALEInterface::setString(const string& key, const string& value) {
   assert(theSettings.get());
   assert(theOSystem.get());
   theSettings->setString(key, value);
   theSettings->validate();
 }
-void ALEInterface::set(const string& key, const int& value) {
+void ALEInterface::setInt(const string& key, const int& value) {
```
mgbellemare: Unless there's a reason, I would pass the ints, bools, and floats by value rather than reference. This lets you call setInt() with a temporary, e.g. the return value of a variable, which you can't do otherwise.

src/environment/stella_environment.cpp
mhauskn: Good point, the std:to_string should be removed.

src/ale_interface.cpp
mhauskn: Yes, good call.

mhauskn:
Changed the PR to remove std::to_string and the set methods to pass by value (except setString which still passes by reference).

PR #42, opened 2015-02-23
Refactored RLGlue Controller and added CMake build support

mhauskn:
Changes:

1) Added Cmake build support for RLGlue Agent + Environment. 
2) rlglue_controller now returns legal_actions by default (or minimal_actions if requested). Before it would just return all 18 actions which could be problematic for games in which some actions are illegal.
3) rlglue_controller now supports _only_ single agent learning. While this is a debatable choice, it goes along with RLGlue's philosophy (http://glue.rl-community.org/wiki/RL-Glue_Core#Does_RL-Glue_support_multi-agent_reinforcement_learning.3F), mirrors the fact that most ALE interest has focused on single-agent learning, and unifies the RLGlue interface (which was previously split between single-agent if -use_restricted_actions was set and multiagent if unset). 

Tested the example RL agent/experiment on Ubuntu 14.04 and verified that it runs successfully.


mhauskn:
One additional thing got pushed into this PR: I added a -sound flag to ALE which toggles the use of sound. Previously sound was enabled whenever SDL was enabled, which caused some weird situations in which no display screen was present but sound was playing.

PR #46, opened 2015-03-01
Goodbye export screen

mgbellemare:
I've added code to crunch out png frames, and finally slain ExportScreen. There are two ways to use this:

1. Either to save a single frame (e.g., via AleInterface::saveScreenPNG()), or
2. To save a sequence of frames, via ScreenExporter::saveNext().

The interface is still a bit clunky, but you can try it out with the following three lines of code:
    #include "common/ScreenExporter.hpp"
    std::auto_ptr<ScreenExporter> exporter(ale.createScreenExporter("/tmp/frame/frame"));
    exporter->saveNext(ale.getScreen());

where the last line should be called after every step of the ALE.


mgbellemare:
I forgot to mention: I'll be adding text to the manual describing how to generate videos via ffmpeg (this is much preferable to trying to add video-making support to ALE).

src/ale_interface.hpp
```diff
@@ -102,6 +103,10 @@ class ALEInterface {
   // Save the current screen as a png file
   void saveScreenPNG(const string& filename);
 
+  // Creates a ScreenExporter object which can be used to save a sequence of frames. Ownership 
+  // said object is passed to the caller.
+  ScreenExporter *createScreenExporter(const std::string &basename) const;
```
mhauskn: What is basename? Describe in comment above.

src/common/ColourPalette.hpp
```diff
@@ -0,0 +1,48 @@
...
+ *   the Reinforcement Learning and Artificial Intelligence Laboratory
+ * Released under the GNU General Public License; see License.txt for details. 
+ *
+ * Based on: Stella  --  "An Atari 2600 VCS Emulator"
+ * Copyright (c) 1995-2007 by Bradford W. Mott and the Stella team
+ *
+ * *****************************************************************************
+ *  ColourPalette.hpp 
+ *
+ *  The colour palette of the now-defunct ExportScreen. Enables conversion from
```
mhauskn: Don't need to mention ExportScreen or the fact that it is defunct.

src/common/ScreenExporter.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <string>
+#include "display_screen.h"
+#include "ale_screen.hpp"
+
+class ScreenExporter {
+
+    public:
+
+        /** Creates a new ScreenExporter which can be used to save screens using save(filename). */ 
+        ScreenExporter(ColourPalette &palette);
```
mhauskn: I assume it takes ColourPalette as an argument to the constructor in case a user wants to use a non-standard palette. I think in most cases, people will just want to use the default palette. In this case it would be nice to have an empty constructor which would just construct a default color palette.

src/common/ScreenExporter.cpp
```diff
@@ -0,0 +1,198 @@
...
+    writePNGData(out, screen, m_palette, true);
+    writePNGEnd(out);
+
+    out.close();
+}
+
+void ScreenExporter::saveNext(const ALEScreen &screen) {
+
+    // MGB: It would be nice here to automagically create paths, but the only way I know of 
+    // doing this cleanly is via boost, which we don't include.
```
mhauskn: Agreed

src/common/ScreenExporter.cpp
```diff
@@ -0,0 +1,198 @@
...
+}
+
+
+ScreenExporter::ScreenExporter(ColourPalette &palette, const std::string &basename):
+    m_palette(palette),
+    m_frame_number(0),
+    m_frame_field_width(6),
+    m_basename(basename) {
+
+    // MGB: Should we strip '.png' from filename if present?
```
mhauskn: Maybe an assert could work here? If a user provides a bad, path it's not going to be able to save anyway, so better just to let them know + fail early?

src/common/ScreenExporter.hpp
```diff
@@ -0,0 +1,58 @@
...
+        ColourPalette &m_palette;
+
+        /** The next frame number. */
+        int m_frame_number;
+
+        /** The width of the frame number when constructing filenames (set to 6). */
+        int m_frame_field_width;
+
+        /** The base path, to which we append the frame number. */
+        std::string m_basename;
```
mhauskn: Maybe we should call this a basepath throughout. Before I read the .cpp file, I was a bit confused that it should be a file path.

mcmachado: Good work Marc,

finally it is going to be easier to record videos.
Are you planning to add features to make it easier to change the screen
being seen (as in your contingency awareness results)? I remember you
mentioning that you have your own implementation of that, and by what I saw
in everybody's pushes, I didn't see that feature.

2015-03-01 12:32 GMT-07:00 Matthew Hausknecht <notifications@github.com>:

> In src/common/ScreenExporter.hpp
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/46#discussion_r25570358>
> :
>
> > +
> > +        /** Save the given screen according to our own internal numbering. */
> > +        void saveNext(const ALEScreen &screen);
> > +
> > +    private:
> > +
> > +        ColourPalette &m_palette;
> > +
> > +        /** The next frame number. */
> > +        int m_frame_number;
> > +
> > +        /** The width of the frame number when constructing filenames (set to 6). */
> > +        int m_frame_field_width;
> > +
> > +        /** The base path, to which we append the frame number. */
> > +        std::string m_basename;
>
> Maybe we should call this a basepath throughout. Before I read the .cpp
> file, I was a bit confused that it should be a file path.
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/46/files#r25570358>
> .
>



-- 
Marlos Cholodovskis Machado

src/environment/stella_environment.cpp
```diff
@@ -212,7 +212,11 @@ const ALEState& StellaEnvironment::getState() const {
 }
 
 void StellaEnvironment::saveScreenPNG(const string& filename) {
-  m_osystem->p_export_screen->save_png(m_screen, filename);
+
```
mhauskn: This seems inefficient. We should have stella env maintain a ScreenExporter as a private member. This member could also be used when fulfilling the m_record_screen_dir flag (see line 139 of stella enviornment.cpp) .

src/common/ColourPalette.hpp
mgbellemare: I guess I got carried away...

src/common/ScreenExporter.hpp
mgbellemare: This is the same old problem of having to pass OSystem around. The colour palette depends on OSystem/Console. Technically we could hardcode it (as NTSC) but I like the ability of supporting SECAM.

I suppose we could have a default ColourPalette which uses a hardcoded NTSC palette. Then ScreenExporter would use said palette. Thoughts?

src/common/ScreenExporter.cpp
mgbellemare: It'll actually all go through fine -- the png writer doesn't depend on the file being named anything reasonable. Related to the other comments on basepath/basename: this is a config pattern I've been using for a little while, where I specify a base path (e.g. /tmp/qbert/frame) and it generates filenames of the form /tmp/qbert/frame######.png . One alternatively is to use a special character (e.g. X, or #) in the base path to specify where the number goes, but I find that a bit clunky. Suggestions welcome.

src/common/ScreenExporter.hpp
mhauskn: After further thought I think it is fine as it is (taking a ColorPalette as an argument). I think we should hide the Screen Exporter behind the Stella Environment since Stella Environment is the one to handle the command line flag for recording games.

src/common/ScreenExporter.hpp
mgbellemare: This only saves pngs to disk. As mentioned, I'll provide instructions to generate a video.

Annotating the images is much more complicated. AleScreen isn't the right data structure for image annotation, since it's in NTSC format (and not RGB). My recommendation is to use CImg, which is header-only. However saving PNGs in CImg is reallly slow. 

Either way, I don't think we should include CImg as part of ALE -- it's a big header file. But maybe one of us could write a wrapper class which lets you use CImg for drawing to an Atari screen, but uses the png writing code.

src/environment/stella_environment.cpp
mgbellemare: The ScreenExporter constructor is actually quite fast, and there's no memory being allocated here -- it's all on the stack. I wouldn't be too concerned about it -- it's only a few cycles compared to the cost of writing a full PNG to disk; and this way it's a bit cleaner.

src/common/ScreenExporter.cpp
mhauskn: Sorry I just meant the png writer will fail though if a bad file path is specified (e.g. /tmp/qbert/ doesnt exist). Okay, so now after your example I see what you mean by the basename/basepath. We should document clearly (maybe with an example) so that others don't get confused like me. I agree with you that the special character (X,#) way is clunky. I'd argue that calling it basepath is more clear to me than basename.

src/common/ScreenExporter.hpp
mgbellemare: There's a bit more flexibility with handing a ScreenExporter to the caller, in the sense that you don't have to record every single frame. For example you might want to record only every k episodes. But maybe that's not needed? 

As Marlos pointed out, ScreenExporter doesn't do everything we might want it to. So it might make sense to restrict its use (e.g. record single frames or all frames), and build a separate visualization suite for more complex operations.

src/ale_interface.cpp
```diff
@@ -254,3 +254,7 @@ void ALEInterface::restoreState(const ALEState& state) {
 void ALEInterface::saveScreenPNG(const string& filename) {
   environment->saveScreenPNG(filename);
 }
+
+ScreenExporter *ALEInterface::createScreenExporter(const std::string &filename) const {
```
mhauskn: Following on my previous comments I'd suggest hiding the ScreenExporter behind StellaEnvironment (and removing the createScreenExporter method). If the user wants to record the whole trajectory, (e.g. the saveNext method of ScreenExporter) they can do it by giving ALE the -record_screen_dir flag and the environment will take care of it. If the want to save individual screens they can use the saveScreenPNG method above.

mhauskn:
Very good job refactoring that pesky export screen into a palette and save screen functionality :)

src/common/ScreenExporter.hpp
mhauskn: Between using the -record_screen_dir flag to record every screen, and the saveScreenPNG function to record single frames, it should be possible to still achieve the same flexibility. I'd argue we should unify or screen saving interface (either to everything ScreenExporter or everything in StellaEnvironment). The only argument for everything in StellaEnv is that it handles the record_screen_dir flag.

mgbellemare: Ahh. Somehow I completely forgot you'd already implemented record_screen_dir, i.e. did all my listed tasks for me. Apologies for stomping on some of your work.

At first I was going to agree with you about letting StellaEnvironment do the saving. But now I'm realizing this doesn't play well with agents that save and load states.

I think we should provide the ScreenExporter object and drop saveScreenPNG from the interface. It might look a bit weird but it encapsulates the functionality better than if it were inside StellaEnvironment, which is about providing a proper RL environment. But I'm open to arguments.

mhauskn: Sorry I got a bit carried away and forgot that the screen recording had been one of your tasks =/

I'm good with dropping saveScreenPNG and just providing the the screenExporter. Since we're on this note, you should take a look at the semantics of the -record_screen_dir flag. It's in the same direction as basename/basepath provided to ScreenExporter, but perhaps the descriptions & naming need to be modified.

mgbellemare:
Okay, I've done the following:
1. Changed ScreenExporter to take in a directory path (rather than partial file path)
2. Changed StellaEnvironment to use ScreenExporter in screen_record_dir mode
3. Dropped saveScreenPNG from StellaEnvironment

Let me know if you see anything else.

mhauskn:
There's still 2 minor comments I had about 1) not needing to mention defunct export screen and 2) explaining basename in ale_interface. Other than that, LGTM!

mgbellemare:
Ah yes -- done now.

mcmachado:
Maybe we should list the remaining tasks. And then decide what to do (like
annotating images).

Also, I'm a little bit concerned people start to use DeepMind's code and
not the ALE. This may be worthy to discuss, because using one or the other
is a huge difference (e.g. they have a "dying" signal).

PR #49, opened 2015-03-10
We're Ready to Test Stochasticity

mhauskn:
This PR adds simple action repeat stochasticity. Additionally it makes display_screen's manual control mode more realistic and fixes one bug with manual control in the ale_interface.


mhauskn:
Merging PR so that fellow UT grad students can do some game testing.

PR #51, opened 2015-04-06
Added Ctypes Python Interface

mhauskn:
Integrated Ben Goodrich's ale_python_inteface (https://github.com/bbitmaster/ale_python_interface). This interface allows python to natively interact with ALE and provides a speed boost when compared to using alternatives such as pipes.

### Example Usage (with cmake):
1) "cmake ."
2) "pip install ." (or "pip install --user ." if you don't have root)
3) "./doc/examples/python_example.py asterix.bin"

### Design rational:
Ctypes allows C (not C++) functions to be called from python. This PR introduces a c-wrapper to ale_interface which provides the functions for ctypes to call. It includes _most_ of the functions in ale_interface with the notable exceptions of those used for loading & saving state. It does include an additional header for parsing the ntsc pallette into RGB. This could potentially be eliminated in favor of the screen exporter.

### Testing:
I have tested python_example.py on Ubuntu and OSX. More extensive testing of functionality would be good.


mhauskn:
I've added the ctypes functions for saveState/loadState.

Displaying screen using SDL and "setBool('display_screen', True)" works on Linux but doesn't work on OSX. On OSX this results in an NSInternalInconsistencyException exception. The exception stems from the main method needing to include SDL. However, since main is called in Ctypes we don't have access to the main method so it is not possible to include SDL/SDLmain. 

Additionally, I've been able to replicate the incorrect-color displays with pygame on OSX. Using the exact same code on Linux results in correct color displays. The integers coming out of the getScreenRGB() are exactly the same on Mac/Linux, but pygame interprets the screen buffer differently. In order to get correct colors on OSX I had to use a byteswap on the numpy array to change the endianness:

```python
    numpy_surface = np.frombuffer(screen.get_buffer(),dtype=np.int32)
    numpy_surface.byteswap(True)
    ale.getScreenRGB(numpy_surface)
    pygame.display.flip()
```

mgbellemare:
Good work! The internet seems to suggest we can call SDL_init directly. It looks like SDL uses the following line:
 #define main SDL_main

So we could either #undef it, or re-define SDL_main to be main again.

Viz pygame colours: this is annoying. I don't anything about pygame. Fiddling around with endianness doesn't seem robust, but maybe we could make it a flag? --fix-colours? That seems safer than writing OS-specific code.

mcmachado:
I think that if we create a troubleshooting section discussing this in the
manual we do not even have to write code, we could just write it on the
manual. I'd be ok with it.

In time, I'm busy these days but in my idle time I am playing with the
python interface, mainly testing performance. I hope I will be able to
start working on my tasks soon.

On 26 April 2015 at 16:51, Marc G. Bellemare <notifications@github.com>
wrote:

> Good work! The internet seems to suggest we can call SDL_init directly. It
> looks like SDL uses the following line:
> #define main SDL_main
>
> So we could either #undef it, or re-define SDL_main to be main again.
>
> Viz pygame colours: this is annoying. I don't anything about pygame.
> Fiddling around with endianness doesn't seem robust, but maybe we could
> make it a flag? --fix-colours? That seems safer than writing OS-specific
> code.
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/51#issuecomment-96443793>
> .
>



-- 
Marlos Cholodovskis Machado

mhauskn:
After a couple hours of painful troubleshooting I'm still coming up short on the SDL_main issue. It doesn't help to call SDL_init -- I tried it and same error resulted. We really just need to call SDL_main(), however I haven't been able to due to linker errors. Some supporting evidence:

In sharedLibraryInterface.cpp I "#undef main" and end up getting the exact same error as in the python example. This is presumably because SDL_main doesn't get called since main is not renamed. 

When I try to add a call to SDL_main(argc, argv) to the sharedLibrary example I get:
```
g++ -D__USE_SDL -DSOUND_SUPPORT -I../../src -I../../src/controllers -I../../src/os_dependent -I../../src/environment -L../.. -I/usr/local/include/SDL -framework Cocoa sharedLibraryInterfaceExample.cpp -lale -lz -lsdl -lsdl_gfx -lsdl_image -lsdlmain -o sharedLibraryInterfaceExample
Undefined symbols for architecture x86_64:
  "_SDL_main", referenced from:
      _main in sharedLibraryInterfaceExample-34b26a.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
So in summary, I haven't been able to get the "#undef main" trick working because of linker errors when manually calling SDL_main and simply calling SDL_Init is not enough.

mhauskn:
Okay, and just after I sent out this comment of despair, I found a hack solution: use pygame to initialize SDL. In the python call:
```python
import pygame
pygame.init()
```
and then everything runs fine! Pretty nasty but maybe we can looks at pygame.init's source and figure out what it's doing to correctly initialize sdl.

mgbellemare:
From a brief glance at the internet (& the pygame documentation) pygame.init seems to
be something that should be called anyway? Not sure why things work without it.

There's also a method called 

pygame.get_sdl_byteorder

which looks like what we're after for the colour issue.

mhauskn:
So there are two separate things going on here:

1) I was hoping to get the basic python example to a point where it could optionally display screen using ALE's internal display without having a dependency on pygame (e.g. the equivalent of sharedLibraryInterfaceExample for python).

2) All the pygame stuff was with reference to a more advanced python example which used the ale_python_interface to get RGB screen colors and then displayed them using pygame. I'm still not sure if we want to include this example (it isn't a part of this PR) because a) it has a dependency on pygame, b) people should prefer ALE's internal display rather than trying to rewrite their own external pygame displays, and c) it has all the nastiness regarding the endianness.

mhauskn:
I've changed the python example to call pygame.init() when on OSX. It's now correctly displaying screen on both OSX and Ubuntu. However, sound does not work on OSX even after the pygame.init call =(.

I don't think we need to include the pygame example as this basic example is capable of displaying screen using ALE's internal screen display.

Admittedly, pygame.init() is a hack, but all things considered, I'm happy with the state of this PR and the state of python support in general. Please test and give feedback :)

mgbellemare:
Works like a charm, nicely done! A few comments:
- Is there any way to avoid having to pip install the package? Not a big deal, but it's an extra step if like me you haven't got pip installed.
- This doesn't work with python3, probably worth documenting somewhere.
- Sound doesn't seem to get output. Is this expected? If so, we should probably set "sound" to false in the example, and comment about it.

mcmachado:
Cool!

I didn't try the display screen yet, but I've already implemented a Sarsa
agent with RAM features and it worked well. I plan to keep implementing
these algorithms to test the Python interface, until the release date.

Finally, FYI (and this is not a criticism, it is just information), my
Sarsa implementation in Python is 6x slower than my C++ implementation. I'm
not a good Python programmer but I don't think I did something terrible
inefficient, but I am still thinking on how to speed up it (*e.g.*, RAM
Features with CTypes). I'm saying this so we can keep this is mind if
someone asks/if we want to use.

But it is very pleasant to implement everything in Python now =)

Cheers.





On 3 May 2015 at 08:54, Marc G. Bellemare <notifications@github.com> wrote:

> Works like a charm, nicely done! A few comments:
>
>    - Is there any way to avoid having to pip install the package? Not a
>    big deal, but it's an extra step if like me you haven't got pip installed.
>    - This doesn't work with python3, probably worth documenting somewhere.
>    - Sound doesn't seem to get output. Is this expected? If so, we should
>    probably set "sound" to false in the example, and comment about it.
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/51#issuecomment-98490228>
> .
>



-- 
Marlos Cholodovskis Machado

mhauskn:
Marc:

The package can be installed using pip or distutils. For distutils I believe the command is 'python setup.py install'. Likely any python programmer will have one of these utilities installed. 

Thanks for testing with python 3. Good to be aware of brokenness. I don't think many people have converted to python 3 yet, but if the error is something simple perhaps we can fix it?

I had the same problem with no sound. I'm unsure of the cause and don't know the solution. I'll set sound to false in the example and document it with a comment in the code. 

Marlos:

Thanks for testing with a Sarsa implementation! It's unfortunate that it's not as fast as the C++ equivalent. I'm curious how much of that results from the interface being slow versus the internals of SARSA not being as fast as their C++ equivalents. 

I did some quick experiments on my laptop by running a game 10 times: It looks like the python interface is slower than C++, but not substantially so. 

Just acting - no getting state:
SharedLibraryExample: 3446 actions/second
Python Example: 3364 actions/second

Getting ram content every frame:
SharedLibraryExample: 3375 actions/second
Python Example: 2982 actions/second

Getting screen every frame:
SharedLibraryExample: 3417 actions/second
Python Example: 2877 actions/second

mcmachado:
Matthew,

you are right. I forgot to mention, but obviously the first test I ran was
only testing the act interface. I tested the performance of the random
agent, without the screen, and I noticed that Python may be 10% slower
(close to yours), maybe less. I'm fine with it.

I'll keep running tests and when I find relevant I'll let you know.

Cheers.

On 3 May 2015 at 17:35, Matthew Hausknecht <notifications@github.com> wrote:

> Marc:
>
> The package can be installed using pip or distutils. For distutils I
> believe the command is 'python setup.py install'. Likely any python
> programmer will have one of these utilities installed.
>
> Thanks for testing with python 3. Good to be aware of brokenness. I don't
> think many people have converted to python 3 yet, but if the error is
> something simple perhaps we can fix it?
>
> I had the same problem with no sound. I'm unsure of the cause and don't
> know the solution. I'll set sound to false in the example and document it
> with a comment in the code.
>
> Marlos:
>
> Thanks for testing with a Sarsa implementation! It's unfortunate that it's
> not as fast as the C++ equivalent. I'm curious how much of that results
> from the interface being slow versus the internals of SARSA not being as
> fast as their C++ equivalents.
>
> I did some quick experiments on my laptop by running a game 10 times: It
> looks like the python interface is slower than C++, but not substantially
> so.
>
> Just acting - no getting state:
> SharedLibraryExample: 3446 actions/second
> Python Example: 3364 actions/second
>
> Getting ram content every frame:
> SharedLibraryExample: 3375 actions/second
> Python Example: 2982 actions/second
>
> Getting screen every frame:
> SharedLibraryExample: 3417 actions/second
> Python Example: 2877 actions/second
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/51#issuecomment-98556369>
> .
>



-- 
Marlos Cholodovskis Machado

PR #55, opened 2015-05-22
Changing the random number generator. Instead of using rand()/srand()…

mcmachado:
**Description:**

I have changed the code so we are now using Mersenne Twister. Now, randomization is something uniform across the ALE. The seed is only set on ale_interface, and it is used by both stochasticity on the control (stella_environment) and on Random.cxx, that feeds the Stella (to initialize the RAM, for example). 

Because I changed the class Random, files like M6532.cxx and Cart*.cxx did not have to be changed.

**Testing:**

I made sure that the seed was going to be set only once at Random.cxx, and when not set, we would use time as default (If we are not cautious, Random is called several times, we could end-up resetting the seed). Therefore, I confirmed that across episodes the seed is kept, and we do not receive the same sequence of random numbers.

I checked we receive the same numbers we would receive if we just implement a simple code that generates random numbers using the same seed with Mersenne Twister.

Finally, I confirmed that when a seed is set, we do not generate one single random number with time (The class Random is first initialized without a seed, so I had to make sure Stella -- or ALE -- does not call next() before setting the seed.)


src/environment/stella_environment.cpp
```diff
@@ -131,7 +131,7 @@ void StellaEnvironment::noopIllegalActions(Action & player_a_action, Action & pl
 
 reward_t StellaEnvironment::act(Action player_a_action, Action player_b_action) {
   
-  double repeat_prob = 0.0;
+  double repeat_prob = 0.25;
```
mhauskn: Ultimately, this value needs to be set from a command line flag.

mhauskn:
Looks good to me!

src/ale_interface.cpp
```diff
@@ -90,14 +90,12 @@ void ALEInterface::loadSettings(const string& romfile,
   // Seed random number generator
   if (theOSystem->settings().getString("random_seed") == "time") {
     cerr << "Random Seed: Time" << endl;
-    srand((unsigned)time(0));
-    //srand48((unsigned)time(0));
+    Random::seed((unsigned)time(0));
```
mgbellemare: time() takes in a pointer as argument. Please change to time(NULL). Also, can you cast to the correct unsigned type (uInt32)? The latter also applies elsewhere in this code.

src/emucore/Console.cxx
```diff
@@ -463,8 +463,8 @@ void Console::initializeAudio()
 */
 void Console::fry() const
 {
-  for (int ZPmem=0; ZPmem<0x100; ZPmem += rand() % 4)
-    mySystem->poke(ZPmem, mySystem->peek(ZPmem) & (uInt8)rand() % 256);
+  for (int ZPmem=0; ZPmem<0x100; ZPmem += randNumGen.next() % 4)
+    mySystem->poke(ZPmem, mySystem->peek(ZPmem) & (uInt8)randNumGen.next() % 256);
```
mgbellemare: Isn't casting next() to uInt8 sufficient to guarantee a modulo 256? I'm not sure what the order of operations is here...

src/emucore/Random.hxx
```diff
@@ -55,8 +56,8 @@ class Random
     uInt32 next();
 
   private:
-    // Indicates the next random number
-    uInt32 myValue;
+    // Mersenne Twister object to randomization
+    static std::mt19937 rndGenerator;
```
mgbellemare: This is a C++11 feature :( It might be nice to have a fallback in the event that someone compiles the ALE with a non-C++11 compiler (they're still around right now, unfortunately).

src/environment/stella_environment.cpp
mgbellemare: Yes, let's change it to a command line flag before we merge.

mgbellemare:
Thanks for the changes, Marlos! I've flagged a few minor issues, but nothing that should take too long to fix. I'm not sure what to do with the C++11 bit... had we agreed we would move to C++11 now?

src/ale_interface.cpp
mcmachado: Ok. I am going to change this.

On 24 May 2015 at 11:03, Marc G. Bellemare <notifications@github.com> wrote:

> In src/ale_interface.cpp
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#discussion_r30953884>
> :
>
> > @@ -90,14 +90,12 @@ void ALEInterface::loadSettings(const string& romfile,
> >    // Seed random number generator
> >    if (theOSystem->settings().getString("random_seed") == "time") {
> >      cerr << "Random Seed: Time" << endl;
> > -    srand((unsigned)time(0));
> > -    //srand48((unsigned)time(0));
> > +    Random::seed((unsigned)time(0));
>
> time() takes in a pointer as argument. Please change to time(NULL). Also,
> can you cast to the correct unsigned type (uInt32)? The latter also applies
> elsewhere in this code.
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55/files#r30953884>
> .
>



-- 
Marlos Cholodovskis Machado

src/emucore/Console.cxx
mcmachado: I don't think so. When I run this code:

#include <iostream>
#include <random>

int main(){
std::mt19937 mt_rand(0);
std::cout << mt_rand() << std::endl;
std::cout << (uint8_t) mt_rand() << std::endl;
std::cout << (uint8_t) mt_rand() % 256 << std::endl;
}

I get the following output:

2357136044
/
117

Therefore, the modulo 256 is necessary. Also, I kept what was before, this
modulo and typecasting was already in the code.

On 24 May 2015 at 11:04, Marc G. Bellemare <notifications@github.com> wrote:

> In src/emucore/Console.cxx
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#discussion_r30953889>
> :
>
> > @@ -463,8 +463,8 @@ void Console::initializeAudio()
> >  */
> >  void Console::fry() const
> >  {
> > -  for (int ZPmem=0; ZPmem<0x100; ZPmem += rand() % 4)
> > -    mySystem->poke(ZPmem, mySystem->peek(ZPmem) & (uInt8)rand() % 256);
> > +  for (int ZPmem=0; ZPmem<0x100; ZPmem += randNumGen.next() % 4)
> > +    mySystem->poke(ZPmem, mySystem->peek(ZPmem) & (uInt8)randNumGen.next() % 256);
>
> Isn't casting next() to uInt8 sufficient to guarantee a modulo 256? I'm
> not sure what the order of operations is here...
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55/files#r30953889>
> .
>



-- 
Marlos Cholodovskis Machado

src/environment/stella_environment.cpp
mcmachado: Ok. I'm going to do it as well.

On 24 May 2015 at 11:06, Marc G. Bellemare <notifications@github.com> wrote:

> In src/environment/stella_environment.cpp
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#discussion_r30953915>
> :
>
> > @@ -131,7 +131,7 @@ void StellaEnvironment::noopIllegalActions(Action & player_a_action, Action & pl
> >
> >  reward_t StellaEnvironment::act(Action player_a_action, Action player_b_action) {
> >
> > -  double repeat_prob = 0.0;
> > +  double repeat_prob = 0.25;
>
> Yes, let's change it to a command line flag before we merge.
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55/files#r30953915>
> .
>



-- 
Marlos Cholodovskis Machado

mcmachado:
Well, I believe we shouldn't have a fallback because this fallback would be
going back to rand/srand. If we do this, we would be adding too much noise
in the experiments, as we will not know what type of randomization was used
in reported experiments. Also, if we set the possibility of using
rand/srand, we would also have the issue of having srand of the agent
overriding the srand of the system (or vice-versa).

I guess we are moving to C++11...

On 24 May 2015 at 11:09, Marc G. Bellemare <notifications@github.com> wrote:

> Thanks for the changes, Marlos! I've flagged a few minor issues, but
> nothing that should take too long to fix. I'm not sure what to do with the
> C++11 bit... had we agreed we would move to C++11 now?
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#issuecomment-105039857>
> .
>



-- 
Marlos Cholodovskis Machado

src/emucore/Console.cxx
mgbellemare: I think what the example is showing is that std::cout prints out uint8_t's as characters (rather than integers) :) Note that the cast has higher precedence than modulo.

But yes, since it was already in the code there's no harm in leaving it there, but it is very strange.

src/emucore/Random.hxx
```diff
@@ -54,11 +55,13 @@ class Random
     */
     uInt32 next();
 
-  private:
-    // Indicates the next random number
-    uInt32 myValue;
+  public:
+    static unsigned long max();
```
mgbellemare: Is there a reason for making this static (when the other methods aren't)?

src/emucore/Settings.cxx
```diff
@@ -300,6 +300,8 @@ void Settings::usage() {
        "     Phosphor blends screens to reduce flicker\n"
        "   -record_screen_dir [save_directory]\n"
        "     Saves game screen images to save_directory\n"
+       "   -stochasticity (default: 0.25)\n"
```
mgbellemare: Can we name this 'repeat_action_probability' instead? We know what "stochasticity" means because we've been talking about it, but it will likely be very cryptic for anyone else.

src/environment/stella_environment.cpp
```diff
@@ -139,12 +142,12 @@ reward_t StellaEnvironment::act(Action player_a_action, Action player_b_action)
   //  past the terminal state
   for (size_t i = 0; i < m_frame_skip; i++) {
     
-    // Stochastically drop actions, according to repeatProb
-    // @todo -- don't use rand()
-    if (rand() / double(RAND_MAX + 1.0) >= repeat_prob)
+    // Stochastically drop actions, according to m_stochasticity
+    if (m_rand_gen.next() / double(MAX_RAND + 1.0) >= m_stochasticity){
```
mgbellemare: Instead of using MAX_RAND, doesn't it make more sense to implement a method Random::nextDouble() which hides away this mess (& guards against bugs)? This is e.g. how Java and the boost Mersenne twister operate.

mgbellemare:
I think it's important to support non-C++11 compilers, at least for a little while longer. This is particularly relevant given that we've already had a number of requests to simplify the installation process. If people need to upgrade their compiler (which can be highly painful on some OS's) then they're likely not to bother -- or simply install an older version. The Mac I'm currently using, for one, didn't come with C++11 installed, even though it's less than 2 years old.

One quick and dirty fix is to add another option to CMake (-DNO_CPP11, or -DUSE_RANDOM_ALTERNATIVE) to fall back to the Random implementation: the old linear congruent one that used to be Random.

This approach raises a small reproducibility issue, as you pointed out, but I think it's a tiny price to pay to make sure people use the ALE.

Thoughts?

src/emucore/Random.hxx
mcmachado: In fact, the method seed() is also static. The unique method that is not
static is next(). I made such method static because it is not dependent of
a particular object, but if you thing I must change this, I'll do it.

On 25 May 2015 at 05:32, Marc G. Bellemare <notifications@github.com> wrote:

> In src/emucore/Random.hxx
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#discussion_r30975437>
> :
>
> > @@ -54,11 +55,13 @@ class Random
> >      */
> >      uInt32 next();
> >
> > -  private:
> > -    // Indicates the next random number
> > -    uInt32 myValue;
> > +  public:
> > +    static unsigned long max();
>
> Is there a reason for making this static (when the other methods aren't)?
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55/files#r30975437>
> .
>



-- 
Marlos Cholodovskis Machado

mcmachado:
Well,

I'd would rather make sure everybody is using the same random number
generator. I do care with reproducibility but, for me the main problem of
using two different implementations is that people may be unaware of the
consequence of their acts if they start using rand/srand with the
option DNO_CPP11.
It is up to you. I would let it the way it is and request people to install
C++11, it is not that hard. But if you want me to add such fallback, let me
know.

On 25 May 2015 at 05:52, Marc G. Bellemare <notifications@github.com> wrote:

> I think it's important to support non-C++11 compilers, at least for a
> little while longer. This is particularly relevant given that we've already
> had a number of requests to simplify the installation process. If people
> need to upgrade their compiler (which can be highly painful on some OS's)
> then they're likely not to bother -- or simply install an older version.
> The Mac I'm currently using, for one, didn't come with C++11 installed,
> even though it's less than 2 years old.
>
> One quick and dirty fix is to add another option to CMake (-DNO_CPP11, or
> -DUSE_RANDOM_ALTERNATIVE) to fall back to the Random implementation: the
> old linear congruent one that used to be Random.
>
> This approach raises a small reproducibility issue, as you pointed out,
> but I think it's a tiny price to pay to make sure people use the ALE.
>
> Thoughts?
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#issuecomment-105217651>
> .
>



-- 
Marlos Cholodovskis Machado

src/emucore/Settings.cxx
mcmachado: I'll change it.

On 25 May 2015 at 05:33, Marc G. Bellemare <notifications@github.com> wrote:

> In src/emucore/Settings.cxx
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#discussion_r30975514>
> :
>
> > @@ -300,6 +300,8 @@ void Settings::usage() {
> >         "     Phosphor blends screens to reduce flicker\n"
> >         "   -record_screen_dir [save_directory]\n"
> >         "     Saves game screen images to save_directory\n"
> > +       "   -stochasticity (default: 0.25)\n"
>
> Can we name this 'repeat_action_probability' instead? We know what
> "stochasticity" means because we've been talking about it, but it will
> likely be very cryptic for anyone else.
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55/files#r30975514>
> .
>



-- 
Marlos Cholodovskis Machado

mhauskn:
I think it's reasonable to require c++0x compatibility at this point in time but too early to require c++11 compatibility. 

While mt19937 is officially C++11, in practice I believe nearly all compliers implemented it as of c++0x. For example I can compile using -std=c++0x without problems. Googling for c++0x random number generators yields multiple results that use mt19937 like this one (https://gist.github.com/opatut/1006641). So in practice I think mt19937 has been around in compilers for quite a while. 

I think we should stick with the Mersenne twister (without requiring c++11) and I'd be surprised if people have compilers that can't handle it.

src/emucore/Random.cxx
```diff
@@ -24,26 +24,32 @@ void Random::seed(uInt32 value)
...
 uInt32 Random::next()
 {
-  return (myValue = (myValue * 2416 + 374441) % 1771875);
+  return rndGenerator();
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+float Random::nextDouble()
+{
+  return rndGenerator() / double(maxPossibleValue + 1.0);
```
mhauskn: Why the + 1.0 here?

mcmachado: I kept it consistent with the previous implementation, where it was being
used RAND_MAX + 1.

On 25 May 2015 at 12:03, Matthew Hausknecht <notifications@github.com>
wrote:

> In src/emucore/Random.cxx
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55#discussion_r30989746>
> :
>
> >  }
> >
> >  // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
> >  uInt32 Random::next()
> >  {
> > -  return (myValue = (myValue * 2416 + 374441) % 1771875);
> > +  return rndGenerator();
> > +}
> > +
> > +// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
> > +float Random::nextDouble()
> > +{
> > +  return rndGenerator() / double(maxPossibleValue + 1.0);
>
> Why the + 1.0 here?
>
> —
> <https://github.com/mgbellemare/Arcade-Learning-Environment/pull/55/files#r30989746>
> .
>



-- 
Marlos Cholodovskis Machado

mgbellemare: Yes, this is to ensure we get a number in the range [0, 1).

mgbellemare:
Re: C++11. I should have mentioned part of my concern is that this branch doesn't compile on this Mac. It's missing

#include <random>

which is afaik a C++11 header. I've updated XCode and haven't quite gotten things to work yet. I'm not running the latest OS X (still on Lion), but I'm concerned that if my upgrade to C++11 is non-trivial / time-consuming then surely others will be deterred by having to upgrade.

This said, let me poke around a bit more and see if this is just a fluke on my end.

mhauskn:
I think your last message was incomplete: did you mean #include <random>?

On Mon, May 25, 2015 at 6:56 PM, Marc G. Bellemare <notifications@github.com

mgbellemare:
Yes.

mgbellemare:
My suggestion was for the fallback to be the Random class that is 
currently in
emucore, not rand/srand. I.e. the one that returns

   return (myValue = (myValue * 2416 + 374441) % 1771875);

Then we can have a big bright message in CMake that tells people to upgrade
to C++0x before the next release.

PR #57, opened 2015-05-24
Dev video recording fix

mgbellemare:
Fixed synchronization issues with sound recording. I had it wrong all along. The current fix is only a 
90% solution. It still requires the SDL display to be running at the right speed. As a result the sound isn't quite as good as it should be. A better solution is to emulate the TIA/sound buffer, i.e. replicate the code
inside SoundSDL and TIASnd to avoid real-time constraints. This will have to wait until ALE 0.6.


mhauskn:
LGTM. Thanks for looking into this Marc!

PR #58, opened 2015-05-30
Replaced std::mt19937 by TinyMT, added relevant paths to CMake/makefile

mgbellemare:
Added TinyMT as a replacement for mt19937 for the time being.


mhauskn:
I think you may have forgotten to include the source/header files for tinymt. Otherwise looks good!

mgbellemare:
Oops, good catch.

PR #68, opened 2015-07-01
fixed wrong namespace inclusion in headers

alcinos:
The file bspf.hxx was declaring "using namespace std". As this file gets included indirectly by the ale_interface.cpp, the whole library was poisoned by the namespace (which is a threat not only internally, but also for the users of the lib)


mhauskn:
Thanks for this PR. It looks good to me. This PR passes required tests (test_ale.sh) on OSX and Ubuntu. I'm in support of merging. I'd prefer to have at least one more pair of eyes before we merge.

mcmachado:
Nicolas, thanks for this PR. Matthew, thanks for looking at it and running the testing script.
I also looked at it and I also ran the required tests. It also looks good to me. Therefore I am merging it.

PR #74, opened 2015-07-08
add verbosity option

alcinos:


CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required (VERSION 2.6)
 project(ale)
 
 option(USE_SDL "Use SDL" OFF)
+option(VEROBSE "Enable log message" ON)
```
mcmachado: There is a typo here. VEROBSE instead of VERBOSE.

alcinos:
Fixed typo pointed out by @mcmachado and added some logging into the OSystem initialization

mcmachado:
I think this is a great feature but wouldn't it be a better if we had it as a flag inside the ALE, and not a compilation flag?

The overhead to check at the start if we are going to print the "verbose" information would be small, and we would still allow people to decide between outputting it or not, depending on their task.

mhauskn:
What is the motivation behind this PR? Was the default logging getting in the way?

mgbellemare:
Thanks @alcinos for the suggested change. Having the option to disable output to stderr is a good idea. One fix needed: rather than macros, let's use a logger object which implements the << operator and conditionally feeds its input to std::cerr. I believe this can be done with something like

```
class Logger {
      template<typename T>
      Logger& operator << (T val) {
          if (enabled) std::cerr << val;
     }
  };
```

Then, as @mcmachado suggested, let's enable this with a command-line option. We could also bring in logging code (e.g. glog), but this seems excessively heavy for such a simple use case.

alcinos:
@mhauskn Yes, having logs that you didn't ask for polluting the output of the program is annoying when you execute scripts on it.

@mcmachado & @mgbellemare : Sure, a run-time configuration is more flexible, the solution I proposed had the only advantage of being extremely simple to implement.

Regarding to a runtime solution, what I can do is to implement the logging class as a purely static class. This is usefull because the classes that will need logging will not necessarily be able to share a common instanciation of the logging class easily. How dois it sound?

To avoid name-polluting of the API, maybe a name like ALELogger is preferable?

mgbellemare:
Yes, a purely static class sounds perfect. Good call on the much better ALELogger. Alternatively feel free to define Logger within the namespace ale -- this is something that we should implement soon anyway.

alcinos:
Sorry for messing the history, the rebasing on master didn't go as smoothly as expected.

I think that the implementation is ready, unless anybody sees something to change. I removed pretty much all std::cerr from the code, except those which were in "print_info" or similar functions, as well as those that checked for the parameter "print_info" before printing (I didn't want a conflict between this parameter and the verbosity parameter.

The usage of the class is pretty straightforward. One can set the verbosity level using for example:
```
ale::Logger::setMode(ale::Logger::Warning);
```
Then only the messages of importance greater or equal than Warning are displayed.

To log, it's easy too:
```
ale::Logger::Info << "this is a log message" << std::endl;
```
The selector at the beginning decides the importance of the message.

src/common/Log.cxx
```diff
@@ -0,0 +1,14 @@
...
+using namespace ale;
+
+Logger::mode Logger::current_mode = Info;
+
+void Logger::setMode(Logger::mode m){
+    current_mode = m;
+}
+
+ale::Logger::mode ale::operator<<(ale::Logger::mode log, std::ostream & (*manip)(std::ostream &)) {
+    manip(std::cerr);
```
mgbellemare: Shouldn't we be checking for current_mode >= mode here?

src/emucore/OSystem.cxx
```diff
@@ -416,10 +420,10 @@ bool OSystem::createConsole(const string& romfile)
   }
   if (mySettings->getBool("display_screen", true)) {
 #ifndef __USE_SDL
-    std::cerr << "Screen display requires directive __USE_SDL to be defined."
-              << " Please recompile with flag '-D__USE_SDL'."
-              << " See makefile for more information."
-              << std::endl;
+    std::ale::Logger::Error << "Screen display requires directive __USE_SDL to be defined."
```
mgbellemare: Should be ale::Logger::Error? :)

src/common/Log.hxx
```diff
@@ -0,0 +1,30 @@
+#ifndef LOG_H
+#define LOG_H
```
mgbellemare: Please use __LOG_HPP__ here (in line with the other header guards).

src/common/Log.hxx
```diff
@@ -0,0 +1,30 @@
...
+{
+    class Logger
+    {
+    public:
+        enum mode{
+            Info = 0,
+            Warning = 1,
+            Error = 2
+        };
+        static void setMode(mode m);
```
mgbellemare: Please put a comment here describing the effect of changing the logger mode. Also, current_mode should probably be private.

mgbellemare:
Aside from my inline comments, LGTM. The returning of a Logger::mode variable is a bit exotic but is fine for our needs.

Note that display_screen.cpp still uses fprintf(stderr, "..."). This isn't logging output per-se, but really something that the user should see when running in interactive mode. Probably these printfs should go to stdout. @mhauskn can weigh in.

We should rename Log.{hxx,cxx} to Log.{hpp,cpp} for consistency (hxx/cxx comes from Stella code). This can be done easily with

    for f in `grep -l -r "Log.hpp"`; do
        sed -i '' s/"Log.hpp"/"Log.hpp"/g $f
    done

from the src directory. I can do this myself after the merge.

alcinos:
I made the requested modifications, except the renaming part.

mgbellemare:
Thanks. Can you also fix the std::ale::Logger bug? The code won't compile without it.

alcinos:
Done too, sorry about that

mgbellemare:
LGTM.

PR #77, opened 2015-07-14
Fix Python getScreenRGB to split color channels.

spragunr:
The existing version of getScreenRGB returns a length 33600 numpy array of 32-bit integers, where the three color channels are embedded as 8-bit fields inside the integer values.  This requires the user to know the bitwise layout of the color channels and write Python code to extract the individual color values. 

This pull request modifies the Python interface so that getScreenRGB returns a height x width x 3 numpy array of type uint8, where each entry represents a single color value in the range 0-255.

### DIFF
--- a/ale_python_interface/ale_c_wrapper.h
+++ b/ale_python_interface/ale_c_wrapper.h
@@ -51,12 +51,20 @@ extern "C" {
   int getRAMSize(ALEInterface *ale){return ale->getRAM().size();}
   int getScreenWidth(ALEInterface *ale){return ale->getScreen().width();}
   int getScreenHeight(ALEInterface *ale){return ale->getScreen().height();}
-  void getScreenRGB(ALEInterface *ale,int *screen_data){
+  void getScreenRGB(ALEInterface *ale, unsigned char *screen_data){
     int w = ale->getScreen().width();
     int h = ale->getScreen().height();
+    int index = 0;
+    int r, g, b;
     pixel_t *ale_screen_data = (pixel_t *)ale->getScreen().getArray();
     for(int i = 0;i < w*h;i++){
-      screen_data[i] = rgb_palette[ale_screen_data[i]];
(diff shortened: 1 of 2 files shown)


mhauskn:
Thanks for this PR! It's super helpful not to have to mess with bit shifting to get the RGB channels.

I've can confirm this works on Ubuntu 14.04. Will test on OSX later today. For reference I used the following modified python example to test that RGB colors are correct:

```python
import sys
from random import randrange
from ale_python_interface import ALEInterface
import matplotlib.pyplot as plt
import numpy as np

if len(sys.argv) < 2:
  print 'Usage:', sys.argv[0], 'rom_file'
  sys.exit()

ale = ALEInterface()

# Get & Set the desired settings
ale.setInt('random_seed', 123)

# Set USE_SDL to true to display the screen. ALE must be compilied
# with SDL enabled for this to work. On OSX, pygame init is used to
# proxy-call SDL_main.
USE_SDL = False
if USE_SDL:
  if sys.platform == 'darwin':
    import pygame
    pygame.init()
    ale.setBool('sound', False) # Sound doesn't work on OSX
  elif sys.platform.startswith('linux'):
    ale.setBool('sound', True)
  ale.setBool('display_screen', True)

# Load the ROM file
ale.loadROM(sys.argv[1])

# Get the list of legal actions
legal_actions = ale.getLegalActionSet()

# Play 10 episodes
for episode in xrange(10):
  total_reward = 0
  while not ale.game_over():
    a = legal_actions[randrange(len(legal_actions))]
    # Apply an action and get the resulting reward
    reward = ale.act(a);
    screen_data = ale.getScreenRGB()
    total_reward += reward
    plt.imshow(screen_data)
    plt.show()
  print 'Episode', episode, 'ended with score:', total_reward
  ale.reset_game()
```

mhauskn:
Followup: same test works fine in OSX. :+1:

mgbellemare:
LGTM.

PR #81, opened 2015-07-22
Change ptrs fix segfault

DoctorTeeth:
This does the following 2 things. 

1. Replace auto_ptr with unique_ptr.
I think it's a good idea to do this now because:
First: auto_ptr has been deprecated, and will actually be deleted in the future, according to Herb Sutter: 
http://herbsutter.com/2014/11/24/updates-to-my-trip-report/
Second: right now it's still a drop-in replacement. unique_ptr has move-only semantics, but you're not doing any copying of auto_ptrs right now. In the future, if you try to do copying of ptrs, the compiler will stop you. If you then still want to do copying, you can switch to shared_ptr. 

2. Instead of segfaulting if someone tries to e.g. get the list of legal actions before a ROM has been loaded, just exit with a message. I think that this is a good change because 
First: segfaulting is always bad
Second: this change would have saved me 5-10 minutes of debugging time, so I expect a similar thing might be true for other people.

### DIFF
--- a/src/ale_interface.cpp
+++ b/src/ale_interface.cpp
@@ -28,7 +28,7 @@
  *  The shared library interface.
  **************************************************************************** */
 #include "ale_interface.hpp"
-
+using namespace std;
 // Display ALE welcome message
 std::string ALEInterface::welcomeMessage() {
   std::ostringstream oss;
@@ -198,7 +198,8 @@ bool ALEInterface::game_over() {
 
 // The remaining number of lives.
 const int ALEInterface::lives() {
-    return romSettings->lives();
+  if (!romSettings.get()){printf("ROM not set\n");exit(1);}
+  return romSettings->lives();
 }
 
 // Applies an action to the game and returns the reward. It is the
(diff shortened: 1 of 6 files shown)


mgbellemare:
Thanks for the fix! I'd like to keep the code C++11 free for just a little bit longer to provide support for certain compilers (MSVC comes to mind). We had an earlier discussion on the topic, however, so unique_ptr will certainly make its appearance in ALE 6-7 months from now.

I'd change the calls to printf to std::cerr (and actually to the logging method that @alcinos is working on). Rather than calling exit(), I would throw an exception, since exit() may not call the appropriate destructors. This also lets the caller recover (somehow).

Since most of your commits are about unique_ptr, I can either do the proposed changes myself & discard this PR, or let you take care of it?

DoctorTeeth:
Very generous of you to support MSVC ;)

I just made a different PR that throws exceptions. It's not clear you need a separate log message in that case, because the user will see the exception.

mgbellemare:
Agreed. Thanks.

PR #82, opened 2015-07-23
Throw exception instead of segfaulting on attempt to acess ROM data before ROM set.

DoctorTeeth:
When you try to, for example, get the number of lives left before loading the ROM, the code should now throw a std::runtime_error.

Optionally, you could also write to a log and mention that this has occurred. 
Seems superfluous - the user will see the exception. 

Optionally, you could move that pointer check into another function to avoid repeating code. 
Also seems a little unnecessary at this point, but I can see the merits.

### DIFF
--- a/src/ale_interface.cpp
+++ b/src/ale_interface.cpp
@@ -198,7 +198,10 @@ bool ALEInterface::game_over() {
 
 // The remaining number of lives.
 const int ALEInterface::lives() {
-    return romSettings->lives();
+  if (!romSettings.get()){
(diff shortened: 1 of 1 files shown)


mgbellemare:
LGTM. Really what should be happening here is ALEInterface should allocate RomSettings & other auxiliary structures within its constructor, so as to avoid the three step process (constructor, set options, loadROM). @mhauskn : we could do this by passing a rom filename + list of options to the ALEInterface constructor. But all of this goes beyond the scope of this PR.

PR #84, opened 2015-08-09
clone/restoreSystemState

### LINKED ISSUE #70: To allow saving and restoring of the random number generator state
I guess a useful feature we can look at adding is to provide a way to save (and restore) the internal emulator state. We do not have even have to support it during an episode, but it would be nice if someone could save the TinyMT state to later restore his job. Right now, if one wants to checkpoint his code, he will have to restart TinyMT.

mgbellemare:
Adds two new methods, clone/restoreSystemState, for explicitly saving the RNG state along with the environment state. Fixes #70


Jragonmiris:
References to these functions should also probably be added to ale_c_interface.

mhauskn:
I've updated the python interface to have these new serialization methods. See pr #96. I've tested basic clone/restore functionality and everything seems to be in order. If it looks good, first merge pr #96 into this branch, and then we can merge this branch into dev.

PR #87, opened 2015-08-10
error: ‘runtime_error’ is not a member of ‘std’

jleni:
A header was missing and I could not build anymore.


mgbellemare:
Oops, good catch.

PR #89, opened 2015-08-10
Adding new methods to python interface: getScreenPaletteXXXX

jleni:
Hi, 
I would like to suggest adding some additional methods to the python interface plus a couple of small changes/improvements.

At the moment, it is common to get RGB data and later to convert to grayscale, etc. This requires applying a 24bits palette in ALE (getScreenRGB) plus a grayscale conversion afterwards. 
With these new methods, it is possible to pass the correct [8 or 24]bits palette, saving CPU time on each frame.

I think this is cleaner/faster than having to use getScreen and applying the palette in python (slow)


mhauskn:
Hi jleni, thanks for the PR. As you point out, ALE doesn't currently have methods for getting a grayscale version of the screen.  We will consider adding this feature in the future if there is enough demand.

As far as I know, the available palettes are NTSC/PAL/SECAM, none of which is grayscale. What palette are you applying in order to get grayscale?

jleni:
Hi Matthew,
The new methods provide only a way to get the frames in any custom palette. It is the script responsibility to decide the palette or create it. The PR is only providing more control to the script.

Anyway, regarding the grayscale example: When a python script starts, it pre-calculates a grayscale palette by calling getPaletteRGB() and doing the conversion itself (Y channel, etc.).
Later, it is only necessary to call ALE with this custom palette to get the frames.

Instead of something like:
````
self.ale.getScreenRGB(self.screen_rgb)
cv2.cvtColor(self.screen_rgb, cv2.COLOR_RGB2GRAY, dst=self.screen_buffer[index, ...])
````

It is possible to do the following:
````
self.ale.getScreenPalette8Bits(self.screen_buffer[index, ...], self.grayscale_palette)
````

In my fork of [deep_q_rl](https://github.com/spragunr/deep_q_rl) I get ~5-15% more frames/sec (my box).
As you say, it can be useful to others that may need PAL/SECAM or even some kind of filtering.

**Comment by mgbellemare** [comment]
@jleni: As always I like the idea of a more efficient ALE. From what I understand your suggestion is to directly do the NTSC -> grayscale conversion, rather than go through RGB. That's a good idea, but I'd like to avoid exposing the notion of colour palette in the interface. I'm not sure there's a strong use case for fast access with arbitrary palettes, especially given the added complexity. My recommendation is to keep getScreenRGB() as it is (rather than split across two methods), and add a getScreenGrayscale() method as needed.

FYI, I believe there is a one to one mapping from NTSC to grayscale: https://en.wikipedia.org/wiki/Grayscale .

**Comment by jleni** [comment]
Thanks for you comments. I reverted my changes and now I provide only a getScreenGrayscale() method. The grayscale conversion was already available in ALE/Stella, though, I had to fix a rounding issue there.

For the sake of completeness, I noticed that the python interface was using an independent hardcoded NTSC palette. PAL/SECAM games were not returning the correct RGB values at the moment. If we want agents to receive rgb/grayscale frames that are independent from the games' display format, this had to be fixed. The c++ ALEInterface is not providing information on the game's display format either…

What I am doing now it getting a copy of the palette buffer every time ALEInterface::loadROM is called. The grayscale palette is calculated by Stella.

Both C++ and python interfaces now provide the correct RGB and grayscale values regardless of the game display format.

File: src/common/ColourPalette.cpp [hunk 1]
```diff
@@ -38,6 +39,10 @@ uInt32 ColourPalette::getRGB(int val) const {
     return m_palette[val];
 }
 
+void ColourPalette::getPaletteCopy(int palette_copy[256])
+{
+    memcpy(palette_copy, m_palette, 256*sizeof(int));
```
mgbellemare: Style suggestion: you can directly write `sizeof(palette_copy)` here to avoid duplicating the constant '256'.

src/emucore/Console.cxx
```diff
@@ -234,6 +234,7 @@ Console::Console(OSystem* osystem, Cartridge* cart, const Properties& props)
 //ALE  myOSystem->eventHandler().allowAllDirections(allow);
 
   myAboutString = buf.str();
+  setColorLossPalette(true);
```
mgbellemare: This seems to be setting the palette to be grayscale by default?

mgbellemare: Ah, I had to dig out the logic again, all good.

mgbellemare:
Thanks, this Python interface looks much better. You're right that we should use the provided Stella palette instead. I'm not sure if round() (your suggestion) or floor() (current setup) is the right operation. My understanding is each NTSC colour (128 of them) should map to a unique grayscale value, so we should double-check that round() results in 128 unique grayscale values. If so, round() sounds like a better choice.

You're right that we should use the Stella palette rather than the hardcoded one (though I'm guilty of the same). I think we can make this code a bit better by moving the palette logic (even for colour, odd for grayscale) into the ColourPalette object. I'm also not sure that we need to copy the palette over: why not simply add a method to ALEInterface which returns the OSystem's palette as a `const int*` ? This can access the OSystem's ColourPalette object, which in turn can return a const int* to its palette. This is also cleaner because it avoids creating a public member variable in ALEInterface. The cost of indirection should be relatively minimal compared to the cost of looping over the screen pixels in `getScreenGrayscale()`.

src/common/ColourPalette.cpp
jleni: Not really... Array decay to pointers when passed as an argument.
If you try sizeof(palette_copy) the result is 8 == the size of int*

We can take the 256 from the declaration but I though it was good as a way to tell in the header that the method expects 256.

jleni:
Rounding:
I compared the results with OpenCV. Without rounding, I noticed that there were a few colors that were not matching... The results now are consistent with OpenCV grayscale conversion.

I didn't want to be too intrusive. If you are happy with exposing m_palette as a const int*, I can easily do that and avoid the copy, duplication, etc. I was not sure if you were going to like it. I agree that the indirection is negligible.

jleni:
So now the palette is not being copied but passed as a const int*

>I think we can make this code a bit better by moving the palette logic (even for colour, odd for grayscale) into the ColourPalette object.

Do you mean moving `setColorLossPalette` functionality to `ColourPalette` ? 
At the moment, `Console` takes care of tracking the correct palette and sends updates to `OSystem`. It would result in bigger changes...

mgbellemare:
ColourPalette already wraps the Stella palette into a friendlier format, so it would make sense to add the new functionality there.

jleni:
The PR is much bigger now but I agree, it looks much better after moving things to ColourPalette

src/ale_interface.cpp
```diff
@@ -101,7 +101,8 @@ void ALEInterface::loadSettings(const string& romfile,
     Random::seed((uInt32)seed);
   }
 
-  theOSystem->console().setPalette("standard");
+  string currentDisplayFormat = theOSystem->console().getFormat();
+  theOSystem->colourPalette().setPalette("standard", currentDisplayFormat );
```
mgbellemare: Spurious whitespace :)

src/common/ColourPalette.cpp
```diff
@@ -38,10 +50,342 @@ uInt32 ColourPalette::getRGB(int val) const {
...
+        paletteFormat = 1;
+    else if (displayFormat.compare(0, 5, "SECAM") == 0)
+        paletteFormat = 2;
+
+    m_palette  = availablePalettes[paletteNum][paletteFormat];
+}
+
+void ColourPalette::calculateGrayscaleValues()
+{
+    for (int type = 0; type < 3; type++)
```
mgbellemare: This a minor point, but this code isn't threadsafe at the moment. Specifically: two threads could currently try to write to currentPalette. It may be worthwhile, at least, to test whether the grayscale values have been initialized, and not set them again. (Also, one could note that these grayscale values aren't going to change, so we could hardcode them).

ale_python_interface/ale_c_wrapper.h
```diff
@@ -51,22 +50,40 @@ extern "C" {
...
+    size_t h = ale->getScreen().height();
+    size_t screen_size = w*h;
+
+    const uInt32 *current_palette = ale->theOSystem->colourPalette().getPalette();
+    pixel_t *ale_screen_data = ale->getScreen().getArray();
+    pixel_t *p = ale_screen_data;
+    pixel_t *q = screen_data;
+
+    for(size_t i = 0; i < screen_size; i++, p++, q++){
+      *q = (unsigned char) (current_palette[*p+1] & 0xFF);
```
mgbellemare: Looking at the new ColourPalette it would be better to directly use ColourPalette.getGrayscale() here, and also ColourPalette.getRGB() above, and avoid exposing the palette itself. But this can be done later.

src/common/ColourPalette.hpp
```diff
@@ -19,6 +19,8 @@
 // Include obscure header file for uInt32 definition
 #include "../emucore/m6502/src/bspf/src/bspf.hxx"
 
+using namespace std;
```
mgbellemare: Let's avoid using namespace std here. Also, missing #include <string>.

src/common/ColourPalette.hpp
```diff
@@ -31,15 +33,53 @@ class ColourPalette {
...
+         *  Calculates grayscale values for all palettes
+         */
+        void calculateGrayscaleValues();
+
+        uInt32 *m_palette;
+
+        bool myUserPaletteDefined;
+
+        // Table of RGB values for NTSC, PAL and SECAM
+        static uInt32 NTSCPalette[256];
```
mgbellemare: One suggestion: I would move these to the cpp file (they don't need to be member variables of ColourPalette) so that they don't clutter the ColourPalette interface.

mgbellemare:
Looks much better now, thanks. There's a few minor things to go over but once that's done I'll merge.

src/common/ColourPalette.cpp
jleni: Well, ALE in general is not thread safe. Checking does not help much unless there is some proper synchronisation mechanism in place... Unless we move to C++ 11 (which I understood you prefer not to), there is no way of making this really thread safe. In addition, every time you load a rom, you need to revise this assignment. If we set m_palette only once, when you load a second rom, you might get incorrect results.

Regarding the grayscale values, yes it could be possible to calculate them once except for the user palettes. I wanted to keep things backwards compatible so kept the, the user palette functionality. I don't know you're opinion about that. Nevertheless, it make take just a few more milliseconds to load the rom but the code is much simpler (unless we totally delete user palettes).

ale_python_interface/ale_c_wrapper.h
jleni: If you think in those terms, I will move the whole thing to ColourPalette. I prefer to loop over pointers in the buffer than calling a method for each pixel. I know I could inline the method it but still. I can create some methods in ColourPalette that deal with buffers instead of single points. What do you think?

src/common/ColourPalette.hpp
jleni: Sure, I will try to make the interface cleaner. Still... 
- loadUserPalette is used outside the class. It is necessary.
- m_palette needs to be a member. Each ColourPalette instance needs each own pointer to the buffer. I guess that at some point, exporting might need to create its own ColourBuffer, etc. this is unless you want to have a singleton global palette, which I am not sure if it makes sense.
- calculateGrayscaleValues(). That would work for standard and z26 but not for user palettes, which are instance related. Again, unless you want the palette object to be a singleton.
- NTSC/etc. I can move them out..

src/common/ColourPalette.hpp
jleni: Well spotted. Changing that.

jleni:
Ok,
- I precalculated the standards/z26 palettes.
- Palettes now are applied inside ColourPalette
- user palettes are not static. When loaded, the grayscale values are calculated.
- Removed many entries from the interface.

jleni:
I have opened another PR for the version number of the python interface. We might want to bump it up in this PR too.
Do you have any preference regarding version numbers? Maybe it should be similar to ALE? 0.5.1.1 / 0.5.1.2?

mgbellemare:
LGTM. ColourPalette::getGrayscale is missing but that can be added easily.

PR #90, opened 2015-08-10
Fix bug, Tetris support, Randomize CPU/Mem 

udibr:
randomization of A,X,Y and Memory is enabled by setting a value to the environment variable ALE_RANDOMIZE_CPU


mhauskn:
Thanks for the PR. It looks like you have included changes to a lot of different aspects of ALE. 

1) The Tetris support looks really cool - can you package this into a PR to ALE's dev branch?
2) Randomizing CPU - Why is this helpful? Why should it be included? Is it better than the fry function in Console.cxx?
3) Did you have trouble with the SDL include on setup.py for OSX? If so please, submit an issue.

udibr:
1) I did a separate pull request fro dev https://github.com/mgbellemare/Arcade-Learning-Environment/pull/92
2) I want to the blocks in Tetris to appear in random order each time. The random mechanisem I wrote needs a different seed. The only entrophy I have is the RAM and registers initial states. Frying function is a little bit crazy to me, it skip some of the memory locations and it does not randomize the registers. In any case there is no way to call the fry after every reset.
3) I did had problems with SDL but eventually I solved them in they way I described. It will have for me to go back and re-create all the issues because I already have a working environment on my only OSX :-)
4) YOU MISSED a bug I fixed in the python wrapper! Do you want that? On which branch?

mhauskn:
Thanks. We had already fixed this bug in dev, but not yet merged to master.

setup.py
```diff
@@ -1,11 +1,20 @@
...
-                    libraries = ['ale_c'],
-                    include_dirs = ['src'],
-                    library_dirs = ['ale_python_interface'],
-                    extra_compile_args=['-D__STDC_CONSTANT_MACROS'],
-                    sources=['ale_python_interface/ale_c_wrapper.cpp'])
+import sys
+if sys.platform == 'darwin':
+    module1 = Extension('ale_python_interface.ale_c_wrapper',
+                        libraries = ['ale_c'],
+                        include_dirs = ['src','/usr/local/include/SDL'],
```
mgbellemare: I didn't have issues compiling the Python interface on Mac OS X. These seem like reasonable flags (except /usr/local/include/SDL is user-specific), but I wouldn't change the setting unless this is better for all users. @mhauskn?

src/games/supported/Tetris.cpp
```diff
@@ -0,0 +1,133 @@
...
+    *rval = *this;
+    return rval;
+}
+
+
+/* process the latest information from ALE */
+void TetrisSettings::step(const System& system) {
+
+    // update the reward
+    int x = readRam(&system, 113);
```
mgbellemare: Can you convert the RAM locations to hex (0x##) for consistency with other files?

src/games/supported/Tetris.hpp
```diff
@@ -0,0 +1,80 @@
...
+        void reset();
+
+        // is end of game
+        bool isTerminal() const;
+
+        // get the most recently observed reward
+        reward_t getReward() const;
+
+        // the rom-name
+        const char* rom() const { return "tetris"; }
```
mgbellemare: Minor comment: possibly this should be called "tetris26", since that's the actual game name.

src/games/supported/Tetris.cpp
```diff
@@ -0,0 +1,133 @@
...
+    int y = readRam(&system, 114);
+    reward_t score = 1 * (x & 0x000F) + 10 * ((x & 0x00F0) >> 4) + 100 * (y & 0x000F);
```
mgbellemare: Can you use getDecimalScore() here? See RomUtils.cpp for details. Also, you're missing the topmost digit (the 1000s).

src/games/supported/Tetris.hpp
udibr: P.S. The original Tetris26.BIN file from 1997 did not randomize the sequence a block which made the game less attractive for ML. I've [added](https://github.com/udibr/tetris26) a pesudo random and along the way decided to rename this to tetris.bin.

Please move comments on Tetris to a sperate [pull request](https://github.com/mgbellemare/Arcade-Learning-Environment/pull/93) I created on the dev branch

src/games/supported/Tetris.hpp
```diff
@@ -0,0 +1,80 @@
+/* *****************************************************************************
+ * The lines 68 and 76 are based on Xitari's code, from Google Inc.
```
mgbellemare: Possibly this bit should be removed from your file.

src/games/supported/Tetris.cpp
udibr: Please move comments on Tetris to a sperate pull request I created on the dev branch

mgbellemare:
I agree with you that a game that always runs the same block sequence is a bit less interesting. Currently the ALE standard is to take games "as they are" (see http://www.arcadelearningenvironment.org/wp-content/uploads/2012/07/bellemare13arcade.pdf). So in this case this means keeping the basic Tetris deterministic (which you have).

I'm in favour of additional stochastic intialization since it clearly helps research. For now the default should be off for reproducibility, but we'll discuss turning it on by default in future releases.

Can you make a separate PR for the randomization code, & merge it to dev rather than master? Thanks.

src/emucore/m6502/src/M6502.cxx
```diff
@@ -81,6 +82,16 @@ void M6502::reset()
   SP = 0xff;
   PS(0x20);
 
+    if (getenv("ALE_RANDOMIZE_CPU")) {
```
mgbellemare: Please name ale_randomize_cpu for consistency. Also set to false in Defaults.cpp. Also indent << 2.

udibr:
OK I will make a separate PR for dev regarding randomization. From my functionality point of view I think the code is OK, but I dont like the way it is currently enabled (i.e. setting environment variable ALE_RANDOMIZE_CPU). What would be a clean way to do that?

src/emucore/m6502/src/M6502.cxx
udibr: OK I will make a separate PR for dev regarding randomization. From my functionality point of view I think the code is OK, but I dont like the way it is currently enabled (i.e. setting environment variable ALE_RANDOMIZE_CPU). What would be a clean way to do that?

mgbellemare:
Can you be more precise -- is your concern about having to explicitly set the ale_randomize_cpu flag?

udibr:
If you think controling behaviour using environment variables is appropriate then OK.
I think I saw there is also a config file and if so then perhaps it should move to there.

mgbellemare:
Right, this shouldn't be a call to get_env -- it should be using Settings. M6502 doesn't currently store a reference to Settings. Note that M6532 already randomizes the RAM, so you may want to check what overlap, if any, there is. My recommendation is to provide a method in both M6532 and M6502 which randomizes the desired bits. Then call it from StellaEnvironment::reset() (via System) so that it's very clear this behaviour is not intrinsic to the emulator, but added for research purposes.

mgbellemare:
One more thing: the randomization process should use OSystem's rng() (see dev branch), to allow clone/restoreSystemState() to guarantee reproducibility. This can possibly wait but if not fixed in your PR, please create an issue.

mgbellemare:
It looks like the extra randomization isn't enough to change Tetris significantly. I'm leaning towards not adding it to the code unless there's a strong use case for it.

mgbellemare:
Unless there's still a desire to merge these changes as they are, I will close the pull request next week.

PR #93, opened 2015-08-12
Tetris support

udibr:
I've added two changes I was missing (whem moving from master to dev)
hope it is working now.

### DIFF
--- a/src/games/supported/Tetris.cpp
+++ b/src/games/supported/Tetris.cpp
@@ -1,0 +1,130 @@
+/* *****************************************************************************
+ * ALE support for Tetris https://github.com/udibr/tetris26
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 2
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


src/games/supported/Tetris.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+/* process the latest information from ALE */
+void TetrisSettings::step(const System& system) {
+
+    // update the reward
+    int x = readRam(&system, 113);
+    int y = readRam(&system, 114);
+    reward_t score = 1 * (x & 0x000F) + 10 * ((x & 0x00F0) >> 4) + 100 * (y & 0x000F);
+    if (score > m_score) {
+        std::cerr << "Score=" << score << std::endl;
```
mgbellemare: Please remove the printout.

mgbellemare:
Tetris support looks good, save for the few comments I made in the other PR (please address).

udibr:
(hopefully) I addressed all the issues regarding Tetris. If you want it to be called Tetris26 and not Tetris then I need to change the class names etc...

mgbellemare:
LGTM. One thing I noticed is that the tetromino sequence is fixed and cyclical. I suspect randomly setting the CPU registers doesn't fix this beyond randomizing where you start in the sequence.

PR #94, opened 2015-08-13
fixing score computation on Atlantis

alcinos:


mgbellemare:
LGTM -- thanks for the fix!

PR #96, opened 2015-08-15
Added clone/restore state methods to ale_python_interface.

mhauskn:
Tested that clone/restore works correctly on Ubuntu. Not sure how to verify the integrity of RNG serialization, but if it's working in the ALE_inteface, it'll be working in the python as well.

### DIFF
--- a/ale_python_interface/ale_python_interface.py
+++ b/ale_python_interface/ale_python_interface.py
@@ -69,6 +69,16 @@ ale_lib.saveState.argtypes = [c_void_p]
 ale_lib.saveState.restype = None
 ale_lib.loadState.argtypes = [c_void_p]
 ale_lib.loadState.restype = None
+ale_lib.cloneState.argtypes = [c_void_p]
+ale_lib.cloneState.restype = c_void_p
+ale_lib.restoreState.argtypes = [c_void_p, c_void_p]
+ale_lib.restoreState.restype = None
+ale_lib.cloneSystemState.argtypes = [c_void_p]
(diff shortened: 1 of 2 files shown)


ale_python_interface/ale_python_interface.py
```diff
@@ -181,13 +190,46 @@ def getRAM(self, ram=None):
...
+        outcomes.  By contrast, see restoreSystemState.
+        """
+        ale_lib.restoreState(self.obj, state)
+
+    def cloneSystemState(self):
+        """This makes a copy of the system & environment state, suitable for
+        serialization. This includes pseudorandomness and so is *not*
+        suitable for planning purposes.
+        """
+        return ale_lib.cloneState(self.obj)
```
mgbellemare: Do you mean to call cloneSystemState here?

ale_python_interface/ale_python_interface.py
```diff
@@ -181,13 +190,46 @@ def getRAM(self, ram=None):
...
+
+    def restoreSystemState(self, state):
+        """Reverse operation of cloneSystemState."""
+        ale_lib.restoreState(self.obj, state)
```
mgbellemare: restoreSystemState?

mhauskn:
Good catches :+1:  I've fixed and tested the (clone/restore)SystemState calls. They seem to work fine, but I didn't verify that RNG is correctly cloned/restored.

mgbellemare:
It is in C++. I'll re-verify before merging dev-rng-serialization-fix.

Issue #78, opened 2015-07-20
Build failing on Windows (msys)


Jragonmiris:
I've been trying to build the ALE on msys. I fixed a few small compilation errors, which you can see by looking at the diffs on my fork:

https://github.com/Jragonmiris/Arcade-Learning-Environment

It mostly just involved explicitly including <time.h> in a couple places, and deleting an argument to mkdir.

However, I'm getting the following linker error for ale.exe

    CMakeFiles/ale-bin.dir/objects.a(ale_interface.cpp.obj):ale_interface.cpp:(.text+0x5cf): undefined reference to `OSystemWin32::OSystemWin32()'
    CMakeFiles/ale-bin.dir/objects.a(ale_interface.cpp.obj):ale_interface.cpp:(.text+0x5fd): undefined reference to `SettingsWin32::SettingsWin32(OSystem*)'
    collect2.exe: error: ld returned 1 exit status
    CMakeFiles/ale-bin.dir/build.make:3509: recipe for target '../ale.exe' failed
    make[2]: *** [../ale.exe] Error 1
    CMakeFiles/Makefile2:60: recipe for target 'CMakeFiles/ale-bin.dir/all' failed
    make[1]: *** [CMakeFiles/ale-bin.dir/all] Error 2
    Makefile:75: recipe for target 'all' failed
    make: *** [all] Error 2

The makefile was generated with -G "MSYS Makefiles". I'm not sure where to proceed from here, perhaps there are some defines somewhere that are being wonky? Because it seems like the object files for the windows systems likely aren't being built.

pkulchenko:
@Jragonmiris, I ran into the same issue and fixed it by changing `if(WINDOWS)` to `if(WINDOWS OR MINGW)` in `CMakeLists.txt`. Otherwise the makefile compiles Unix/POSIX version instead of the Windows one and that's why the linking step fails.

PR #97, opened 2015-08-15
Updated to fix compilation using mingw on Windows.

pkulchenko:
This fixes mingw/gcc compilation issues for me. Related ticket #78.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ if(APPLE)
   set(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
 endif()
 
-if(WINDOWS)
+if(WINDOWS OR MINGW)
   list(APPEND SOURCES ${SOURCE_DIR}/os_dependent/SettingsWin32.cxx ${SOURCE_DIR}/os_dependent/OSystemWin32.cxx ${SOURCE_DIR}/os_dependent/FSNodeWin32.cxx)
 else()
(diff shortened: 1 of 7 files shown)


src/os_dependent/FSNodePOSIX.cxx
```diff
@@ -230,5 +230,10 @@ bool AbstractFilesystemNode::dirExists(const string& path)
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 bool AbstractFilesystemNode::makeDir(const string& path)
 {
-  return mkdir(path.c_str(), 0777) == 0;
+  return mkdir(path.c_str()
```
mgbellemare: I believe `FSNodePOSIX.cxx` shouldn't be compiled under Windows, and looking at your change to CMakeLists.txt that should also be true for mingw. Can you verify that this edit is still necessary?

pkulchenko:
@mgbellemare, you are correct, the change in `FSNodePOSIX.cxx` is no longer needed. I tested with the latest master and it works as expected. Rebased and pushed.

mgbellemare:
LGTM. Thanks for the fix.

Jragonmiris:
Is there any way this can also be routed into the dev branch somehow? I'm trying to use the dev branch, but this is only on master and it's a bit wonky since I get to basically choose whether it compiles or uses new features.

pkulchenko:
I wasn't even aware there was a dev branch. @mgbellemare, what's the difference between the dev and the master branches?

mgbellemare:
We've been using the dev branch for tagged releases. A number of improvements made their way directly into master, with the unfortunate side-effect of having slightly different features in both. We're hoping to merge this weekend, so everything should be sorted then.

PR #98, opened 2015-08-16
Compilation problem: Missing files Log.hpp/cpp

jleni:
After commit 6a3e061 some files were missing [Log.hpp/cpp]. Taking them from the previous commit and renaming accordingly.


mgbellemare:
Thanks for catching this quickly.

PR #100, opened 2015-08-20
Adding a version number to the python interface

jleni:
The python interface doesn't have a version number. The current default value is 0.0.0
pip doesn't upgrade and it has to be forced through --upgrade. In addition, it is not clear which version is being used.
I am not sure how you would like to treat version numbers so I am temporarily adding version=0.0.1


mgbellemare:
I'm not sure whether the Python interface should have the same version number as the ALE (though that is reasonable). @mhauskn can weigh in on the issue? For now, merging your change to allow for the new fixes.

mhauskn:
It would be confusing to have different version numbers for ALE and python interface. Lets keep them in sync.

mhauskn:
If we needed to make small changes to the python interface at a faster pace than new ale versions arrive, we could just add another decimal point to its version.

mgbellemare:
Sounds good to me.

PR #102, opened 2015-09-11
MinimalActions - Breakout / Pong

jleni:
Hi,
I've been using Nathan Sprague's implementation (deep_q_rl) and I have struggled to achieve results that are consistent with DQN (Deepmind).

I recently noticed that ALE is different from Xitari in several games. In particular, Breakout and Pong declare a different set of minimal actions.

In this PR, I am removing some spurious actions only in Breakout/Pong in order to match Xitari.

I've seen some differences in other games. Maybe you can comment on that. Should we fix those too? Are Atlantis/Frostbite better?

|Game|ALE|Xitari|
|---|---|---|
|Asteroids| | Score wrapping|
|Atlantis|Seems newer/better||
|Breakout|6 actions|4 actions|
|Frostbite|Seems newer/better||
|Kangaroo||Score is taken from different memory regions|
|Pong|6 actions|3 actions|
|Space Invaders|Score wrapping||
|Wizard of Wor||Different bytes for score/lives|
|Zaxxon||Different mask for lives|

I wonder how these differences affect publications that have been using ALE and benchmarking against DQN/Xitari.

### DIFF
--- a/src/games/supported/Breakout.cpp
+++ b/src/games/supported/Breakout.cpp
@@ -84,8 +84,6 @@ bool BreakoutSettings::isMinimal(const Action &a) const {
         case PLAYER_A_FIRE:
         case PLAYER_A_RIGHT:
         case PLAYER_A_LEFT:
-        case PLAYER_A_RIGHTFIRE:
-        case PLAYER_A_LEFTFIRE:
             return true;
         default:
             return false;
--- a/src/games/supported/Pong.cpp
+++ b/src/games/supported/Pong.cpp
@@ -64,11 +64,8 @@ bool PongSettings::isMinimal(const Action &a) const {
 
     switch (a) {
         case PLAYER_A_NOOP:
-        case PLAYER_A_FIRE:
         case PLAYER_A_RIGHT:
         case PLAYER_A_LEFT:
-        case PLAYER_A_RIGHTFIRE:
-        case PLAYER_A_LEFTFIRE:
             return true;
         default:
             return false;


mgbellemare:
Thanks for the PR -- I'll take a closer look over the weekend. Xitari is not being updated so it is inevitably going to diverge from the ALE. Save for the games where the score is defined differently, reproducibility should not be an issue; this said, I would be very curious to see the actual numbers if you have them handy. Does going from 6 to 4 actions in breakout significantly change the shape of learning curve?

jleni:
I guess that that there are still other reasons why deep_q_rl is still below par but this was a finding that I though was good to notice and fix.

Regarding DQN, it would be necessary to collect more data. 
I ran breakout with DQN (clean) and DQN with Xitari patched to use 6 actions (only 40 epochs.. it is still running)

The effect seems minimal if there is an effect at all (a tiny bit lower avg. score (epoch30-40) but it could be just chance. What do you think?)

Still, I assume that there might be are other approaches are that are more sensitive to a higher branching factor.

By the way, a little off-topic, maybe we can take it offline if you have the time:
Are you aware of any publication where the use of the second network in DQN is analysed/justified?


DQN-Breakout-4Actions:


DQN-Breakout-6Actions:

mgbellemare:
Answering your more general question: Atlantis, Frostbite, and Zaxxon contain bug fixes that are not present in Xitari. I added the Asteroids score wrapping code -- best would be for someone to do this on a per-game basis. See issue #103 which I've just opened on the topic. Wizard of Wor and Kangaroo are the same: there is equivalence in the memory accesses: the 8th bit is not used, but convention sets is to 1, hence 0x86 instead of 0x06.

mgbellemare:
Yes, email me if you have research-specific questions. The smaller branching factor certainly made a difference in the UCT results in the original JAIR paper (Bellemare et al., 2013). There we didn't have a minimal action set defined but we directly compared the RAM states to figure these out. I think DQN should be robust enough to minimally suffer from extra actions, but it would be interesting to see a deeper analysis.

jleni:
Thank you very much for looking at this. 
I really appreciate your work.

spragunr:
Anecdotally, I can report that including irrelevant actions has a surprisingly small impact on performance in DQN.  My original implementation always used the full action set for all games.  I expected big improvements when I moved to reduce action sets, but the improvements turned out to be relatively modest.

Issue #83, opened 2015-08-08
Allow game state serialization from C interface


Jragonmiris:
At the moment, the C interface allows saving and loading of the game state onto an internal, opaque stack. It would be useful to be able to explicitly tell it to save/load a state from disk. 

I've been running UCT from an FFI wrapper, which can take several days, and it would be nice to be able to halt my program and restart my computer (or make it robust against power outages or whatever).

I'd guess this probably isn't much more complicated than saving the RAM query and a few small values like lives and score from the game-specific glue, but the wrapper lacks any real way to load this even if you can grab the values.

mgbellemare:
This is actually easy to implement, since as you point out the save/load state process is already serializing nearly everything to a string. My recommendation is to add a pair of public methods to ALEState, `save(const char*)` and `load(const char*)` which write/read all of ALEState's private member variables to disk. Then you can do

    ALEState s = my_ale_interface.clone();
    s.save("/tmp/my_checkpoint");

mgbellemare:
Also, see issue #70 and consider submitting the PR to the dev branch if possible.

Jragonmiris:
Another idea, which I've implemented, is simply to serialize ALEState to a string. So you have (in ALEState):

    ALEState(std::string serialized);
    std::string encode();

The code for this is simple, it's just the fields of ALEState in declaration order separated by newlines. The parsing is just `getline` and `atoi` (`atoi` instead of `stoi` because unfortunately `stoi` doesn't exist on MinGW), with the final getline reading directly into the string field.

I prefer this to an explicit save to disk method because it allows wrappers from other languages to choose how to save the string -- naked file, with other structs in their own program, gzip it, etc. It's more compatible with other APIs.

Jragonmiris:
The C interface would then just have an `char *aleStateEncode(ALEState *s);` and the reverse, which you can save/load with the new `saveSystemState` methods and such.

mhauskn:
As a side note: I had attempted something similar to this when adding the current save/restore methods to the C interface. I had tried to return the ALEState::m_serialized_state as a char *, but found out that it quite often contains null terminators which resulted in char * being prematurely terminated.

Jragonmiris:
This abuses a lot of C, but this code seems to work as long as you use the char * as a raw buffer of bytes instead of a c string:  http://cpp.sh/3hqc

Jragonmiris:
[...]

PR #106, opened 2015-09-30
Added ALEState serialization

Jragonmiris:
This adds the ability to serialize an `ALEState` to a `vector<char>`, and exposes this as a `char *` in the C interface (with explicit documentation that this is NOT a C-style string). This should address #83 by allowing any user accessing the ALE via the C wrapper or linking directly with the C++ library to grab the raw bits and save them however they please by first performing a `cloneState` or `cloneSystemState`.

The first commit (105623a) is a clone of #97 applied to the dev branch, which I can remove if that commit can somehow be cherry picked into the dev branch. Without it I can't compile this branch on Windows which makes testing this code difficult. This should probably be done before it's merged to avoid merge conflicts when `dev` gets merged with `master`.

PTAL

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ if(APPLE)
   set(CMAKE_SHARED_LIBRARY_SUFFIX ".so")
 endif()
 
-if(WINDOWS)
+if(WINDOWS OR MINGW)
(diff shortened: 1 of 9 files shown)


Jragonmiris:
Note that this method isn't, strictly, portable in that if you take the char vector and send it between two computers with different endianness it will probably break, but I think it's probably "good enough" for our purposes unless we want to go crazy and use boost for serialization or some nonsense.

mgbellemare:
Re: my comment in issue #83 . I would prefer to use the existing Serializer/Deserializer code unless these are insufficient. We're hoping to merge the dev branch within a day or two, so maybe it also makes sense to resubmit this PR then?

PR #107, opened 2015-09-30
Verification before setting flags in the ALE (Issue 76)

mcmachado:
This PR addresses the Issue 76. The rationale is: in the methods `setInt`, `setFloat`, `setBool`, and `setString`, before calling `setExternal`, we now verify whether the variable being set exists or not (this happens in the method `verifyVariableExistence`). This method checks if the variable being set is defined as valid (it works with `templates`, because now each variable has a type). I made several changes on how flags are set, I'll describe them below:

* I deleted the files `Default.xxx`. Now all the flags are being set in the same place, `Settings.cxx` and its default values are also there.
* I created a `map` for each type of flags one can set, namely `intSettings`, `boolSettings`, `floatSettings`, and `stringSettings`.  I did not use `unordered_map` because it is a C++11 feature and we do not support it yet. It should be trivial to change, if desired, when we start to support it. Also, I decided to define different types for each variable to make it more transparent and to make sure when someone calls `setBool("repeat_action_probability", false)`, for example, it does not work.
* In the past, the default value for the variable `random_seed` was `"time"`. However, because `random_seed` only accepts integers now, its default value is `-1` and I properly changed the seed initialization in `OSystem.cxx` as well.
* In the past, the default values were being set in `ale_interface.cpp`. I think this should be more transparent, having them set in the `Settings` constructor. They are now set this way. I didn't see a place where this could have a negative effect (or introduce a bug).

I tested this new implementation and it seems to work fine. An important remark is that, if I forgot to add any existent variable as default (`setDefaultSettings`), when someone tries to set this variable it will fail. I made sure the examples `videoRecordingExample` and `sharedLibraryInterfaceExample` work the same way as before.


src/emucore/OSystem.cxx
```diff
@@ -183,7 +183,7 @@ void OSystem::resetRNGSeed() {
 
   // We seed the random number generator. The 'time' seed is somewhat redundant, since the
   // rng defaults to time. But we'll do it anyway.
-  if (mySettings->getString("random_seed") == "time") {
+  if (mySettings->getInt("random_seed") == -1) {
```
mgbellemare: Can we make this 0 instead of -1?

src/emucore/Settings.cxx
```diff
@@ -27,6 +27,8 @@
 #include "bspf.hxx"
 #include "Settings.hxx"
 
+using namespace std;
```
mgbellemare: I would discourage a using directive for a single class (ostringstream), but this is fine as is.

src/emucore/Settings.cxx
```diff
@@ -92,6 +94,8 @@ Settings::Settings(OSystem* osystem) : myOSystem(osystem) {
     setInternal("uipalette", "0");
     setInternal("mwheel", "4");
     setInternal("autoslot", "false");
+
+    this->setDefaultSettings();
```
mgbellemare: No this->.

src/emucore/Settings.cxx
```diff
@@ -672,3 +688,59 @@ Settings& Settings::operator = (const Settings&)
 
   return *this;
 }
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+void Settings::setDefaultSettings() {
+
+    // Stella settings
+    stringSettings.insert(pair<string, string>("cpu", "low")); // Reduce CPU emulation fidelity for speed
```
mgbellemare: Having one container for each setting type is a bit clunky, but it is a nice safe option. We could consider hiding the multiplicity of containers behind a templated method which automatically returns the correct container, but this will do too.

src/emucore/Settings.cxx
```diff
@@ -672,3 +688,59 @@ Settings& Settings::operator = (const Settings&)
...
+    for(map<string, int>::iterator it = intSettings.begin(); it != intSettings.end(); it++) {
+      this->setInt(it->first, it->second);
+    }
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+template<typename TYPE>
+void Settings::verifyVariableExistence(map<string, TYPE> dict, string key){
+    if(dict.find(key) == dict.end()){
+      cerr << "The key " << key << " you are trying to set does not exist." << endl;
```
mgbellemare: Throw an exception here, rather than calling exit().

src/emucore/Settings.cxx
```diff
@@ -672,3 +688,59 @@ Settings& Settings::operator = (const Settings&)
...
+      this->setBool(it->first, it->second);
+    }
+
+    for(map<string, int>::iterator it = intSettings.begin(); it != intSettings.end(); it++) {
+      this->setInt(it->first, it->second);
+    }
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+template<typename TYPE>
```
mgbellemare: I would use regular capitalization, e.g., ValueType, over TYPE which suggests a constant value.

src/emucore/Settings.hxx
```diff
@@ -162,6 +164,9 @@ class Settings
     */
     void setSize(const string& key, const int value1, const int value2);
 
+    /** Sets all of the ALE-specific default settings */
+    void setDefaultSettings();
```
mgbellemare: Do we want this method to be callable anywhere but from the constructor? If not, I would make it private.

mgbellemare:
Some cosmetic comments as a whole looks good. This is a much-needed change. We need to document the rng change in the manual as well -- it might be easier if you do it in this PR if it isn't too much work. I'll go through some testing rounds now.

src/emucore/OSystem.cxx
mcmachado: Sure. I'll change it now. Since it is not an unsigned int, I don't see a difference, but it is fine by me if you want it to be 0.

src/emucore/Settings.cxx
mcmachado: Ok. I'll put the std:: instead of using namespace.

src/emucore/Settings.cxx
mcmachado: It is not clear for me. Do you want me to remove the "this->"?

src/emucore/Settings.cxx
mcmachado: Ok. I'll change it here.

src/emucore/Settings.cxx
mcmachado: Sure. Since we never agreed on what to do with errors, I used the same pattern that is in this file when someone tries to get a variable that does not exist.

src/emucore/Settings.cxx
mcmachado: Ok. I'll leave it as it is for now.

src/emucore/Settings.hxx
mcmachado: Ok. I thought one could want to reset the system from somewhere else, but it is a silly thing. Private makes much more sense.

mcmachado:
I just pushed the changes you suggested, @mgbellemare. Please let me know if there is something else.

mgbellemare:
Settings.cxx is missing

`#include <stdexcept>`

otherwise all good.

PR #109, opened 2015-10-05
Serialization of ALEStates to strings

Jragonmiris:
This is on top of the new head of dev after master was merged into it. It also addressed the comments made in #106 and #83, and uses the Stella serializer/deserializer which I didn't know existed.

### DIFF
--- a/ale_python_interface/ale_c_wrapper.cpp
+++ b/ale_python_interface/ale_c_wrapper.cpp
@@ -2,9 +2,14 @@
 
 #include <cstring>
 #include <string>
+#include <stdexcept>
 
-void encodeState(ALEState *state, char *buf) {
+void encodeState(ALEState *state, char *buf, int buf_len) {
 	std::string str = state->serialize();
+
+	if (buf_len < int(str.length())) {
+		throw new std::runtime_error("Buffer is not big enough to hold serialized ALEState. Please use encodeStateLen to determine the correct buffer size");
+	}
 
 	memcpy(buf, str.data(), str.length());
 }
--- a/ale_python_interface/ale_c_wrapper.cpp
+++ b/ale_python_interface/ale_c_wrapper.cpp
@@ -1,3 +1,20 @@
-//all code is currently in the .h file
 #include "ale_c_wrapper.h"
 
+#include <cstring>
+#include <string>
+
+void encodeState(ALEState *state, char *buf) {
+	std::string str = state->serialize();
+
+	memcpy(buf, str.data(), str.length());
+}
+
+int encodeStateLen(ALEState *state) {
+	return state->serialize().length();
+}
+
+ALEState *decodeState(const char *serialized, int len) {
+	std::string str(serialized, len);
+
+	return new ALEState(str);
+}
--- a/ale_python_interface/ale_c_wrapper.h
+++ b/ale_python_interface/ale_c_wrapper.h
@@ -77,6 +77,10 @@ extern "C" {
   void restoreSystemState(ALEInterface *ale, ALEState* state){ale->restoreSystemState(*state);}
   void deleteState(ALEState* state){delete state;}
   void saveScreenPNG(ALEInterface *ale,const char *filename){ale->saveScreenPNG(filename);}
+
+  void encodeState(ALEState *state, char *buf);
+  int encodeStateLen(ALEState *state);
+  ALEState *decodeState(const char *serialized, int len);
 }
 
 #endif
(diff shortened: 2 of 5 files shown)


Jragonmiris:
And I should mentioned that I've tested this and it is working properly.

ale_python_interface/ale_c_wrapper.cpp
```diff
@@ -1,3 +1,20 @@
-//all code is currently in the .h file
 #include "ale_c_wrapper.h"
 
+#include <cstring>
+#include <string>
+
+void encodeState(ALEState *state, char *buf) {
+	std::string str = state->serialize();
+
+	memcpy(buf, str.data(), str.length());
```
mgbellemare: What's the rationale for using char *buf instead of std::vector, or simply std::string? In particular, this current form is unsafe since buf may not be of sufficient size.

ale_python_interface/ale_c_wrapper.cpp
```diff
@@ -1,3 +1,20 @@
...
+}
+
+int encodeStateLen(ALEState *state) {
```
mgbellemare: It seems to me like we could do away with this method if encodeState directly returned a dynamically resizing object.

src/environment/ale_state.cpp
```diff
@@ -33,6 +33,16 @@ ALEState::ALEState(const ALEState &rhs, std::string serialized):
   m_serialized_state(serialized) {
 }
 
+ALEState::ALEState(std::string serialized) {
```
mgbellemare: Can pass by const & to avoid a spurious copy.

mgbellemare:
Looks mostly good, save for the char* buf bit. Once that's taken care of I'll merge.

ale_python_interface/ale_c_wrapper.h
```diff
@@ -77,6 +77,13 @@ extern "C" {
   void restoreSystemState(ALEInterface *ale, ALEState* state){ale->restoreSystemState(*state);}
   void deleteState(ALEState* state){delete state;}
   void saveScreenPNG(ALEInterface *ale,const char *filename){ale->saveScreenPNG(filename);}
+
+  // Encodes the state as a raw bytestream. This may have multiple '\0' characters
+  // and thus should not be treated as a C string. Use encodeStateLen to find the length
+  // of the buffer to pass in, or it will be overrun as this simply memcpys bytes into the buffer.
+  void encodeState(ALEState *state, char *buf);
```
mgbellemare: Did you test this in Python? The `ale_python_interface.py` wrappers seem to be missing.

ale_python_interface/ale_c_wrapper.cpp
Jragonmiris: Because that's the C wrapper, and C convention is generally to pass in a target buf. It wouldn't work with FFI if you had to pass in a vector.

ale_python_interface/ale_c_wrapper.cpp
Jragonmiris: That's true, but this way allows wrappers to avoid dynamic allocations if they want.

ale_python_interface/ale_c_wrapper.h
Jragonmiris: No, I didn't. I was using my [own wrapper for Rust](https://github.com/Jragonmiris/ale). I'll add the python bindings.

ale_python_interface/ale_c_wrapper.cpp
mgbellemare: Ah, right. Can you also pass in the buffer size for safety, & throw an exception as needed?

ale_python_interface/ale_c_wrapper.cpp
mgbellemare: If it isn't using C++, then no, it's all good.

ale_python_interface/ale_c_wrapper.cpp
Jragonmiris: Yeah, sure.

Jragonmiris: I'd like to point out that getRAM and getScreen and such don't take in a safety length, should I add that for those too to keep it uniform?

mgbellemare: Good catch. I'd like to merge into master & tag asap, so we can also just make it an issue for now, & fix it after releasing 0.5.1. @mhauskn who may want to weigh in on this.

Jragonmiris:
All comments should have been addressed.

mgbellemare:
LGTM.

mhauskn:
I think the newer method of throwing an error for incorrect buffer size is
a good idea. On the python side, many of the getScreen/Ram/etc methods
allow the user to pass in their own buffer -- which may or may not be
correctly sized. Throwing an error would be more helpful than a segfault.

In general, for each of these methods that does a memcpy, I think we should
have an accompanying getSize method which returns the length the memory
buffer should be.

Issue #111, opened 2015-10-07
Inconsistency between Python interface and C++ interface


mcmachado:
The methods `getScreenGrayscale` and `getScreenRGB` are available in the Python interface but they are not available in the C++ interface (`ale_interface.*pp`). I think both interfaces should be consistent.

mcmachado:
I'll suggest a solution today.

mgbellemare:
Fixed by #113.

PR #113, opened 2015-10-07
Adding functions getScreenRGB and getScreenGrayscale.

mcmachado:
I'm just adding two new methods that were implemented in the Python interface but that were not available in the C++ interface. This is supposed to address the issue #111. I was not sure the best interface for these methods, I guess one could implement a class ScreenRGB and another class, ScreenGrayscale, so we could return objects from these classes, but at a first moment I thought this would be too convoluted. A simpler solution is to expect the user to provide an array (I tried to avoid managing the array memory, so I rely on the user) and we fill that array, the same way it works on the Python interface. The PR I created for the dev branch describes the interface I am suggesting.

To use the method, a simple code like that is enough:

            `int screenSize = 3 * ale.getScreen().width() * ale.getScreen().height();
            pixel_t rgb[screenSize];

            ale.getScreenRGB(rgb);`

I'm pretty sure there are better interfaces to be proposed, but this interface consists in the minimum effort. I think it is reasonable.


mgbellemare:
We should use dynamically allocated memory here, preferably std::vector. But we should also consider using or extending ALEScreen, which already encapsulates an Atari 2600 screen. Do you think this change is necessary for 0.5.1? If not, we could consider postponing the change until we've had time to think about the right encapsulation.

mcmachado:
I don't think it is absolutely necessary. I just fell bad that people using the Python interface will have two useful functions that are not easily available in the C++ interface. I proposed an easy solution expecting that we would change it soon. If you don't want to release the version 0.5.1 with the easy "fix", it is up to you.

mgbellemare:
Okay, let's wait then.

jleni:
Sorry to jump in but I find this useful. 
How about passing by reference an std::array so you can be sure that the size is always correct?

jleni:
I completely understand if it is too late for this release. No worries

src/ale_interface.hpp
```diff
@@ -107,6 +107,17 @@ class ALEInterface {
   // Returns the current game screen
   const ALEScreen &getScreen();
 
+  //This method should receive an array of length width x height
+  //(generally 160 x 210 = 33,600) and then it will fill this array
+  //with the grayscale colours
+  void getScreenGrayscale(pixel_t *grayscale_output_buffer);
+
+  //This method should receive an array of length 3 x width x height
```
mgbellemare: The comment needs to be updated: this is [currently computed as] width x height x 3, rather than 3 x width x height.

mgbellemare:
Returning to this PR now. @jleni: std:array would work in most cases, but some games have nonstandard screen sizes. @mcmachado : can you convert the getScreenRGB and getScreenGrayscale functions to take a std::vector& as input? The better solution is to have a proper encapsulation (i.e. like an ALEScreen, which contains width & height information) but this solution is a bit more work.

mcmachado:
@mgbellemare I will do it. I'll change the getScreenRGB and getScreenGrayscale to take a std::vector& as input.

mhauskn:
I'd also like to use the getScreenGrayscale method. Can we finish + merge this PR?

mcmachado:
@mgbellemare I implemented a version of it, but before PR I wanted to double check the best solution.

1) I had to decide between using the already implemented `applyPaletteGrayscale`/ `applyPaletteRGB` methods and then copy from the `uInt8* dst_buffer` to the `vector<unsigned char>` after the conversion is done; or to re-implement `applyPaletteGrayscale`/ `applyPaletteRGB` receiving `vector<unsigned char>` as parameter. I chose the later. The good thing is that it is faster (I do not have to copy the whole screen one extra time), the bad thing is maintenance. If such a function is changed it has to be changed in two places.

2) I had to decide between expecting the user to provide a vector with the correct size of if I was going to fill the vector every step. For efficiency I decided to expect the user is going to provide us the vector with the correct number of positions. This avoids frequent memory allocations.

3) I decided to "return" (void function) a `vector<unsigned char>` for the grayScale with 210 x 160 positions filled. For the RGB colors I return  `vector<vector<unsigned char> >`, where the first vector has size 3 (R, G and B) and the inner vector has size 210 x 160.

Are you ok with these decisions?

mcmachado:
Small fix: To make it more similar to the python interface I'd suggest we have an RGB array where we access the elements as `rgb[pixel][0]`, where the second dimension is R (0), G(1) or B(2). But I think we should not split between rows and columns. What do you think?

mcmachado:
@mgbellemare I just pushed a new version. Please check if you are ok with it.

src/ale_interface.cpp
```diff
@@ -251,6 +251,36 @@ const ALEScreen& ALEInterface::getScreen() {
...
 }
 
+//This method should receive an empty vector to fill it with
+//the grayscale colours
+void ALEInterface::getScreenGrayscale(std::vector<unsigned char>& grayscale_output_buffer){
+  size_t w = environment->getScreen().width();
+  size_t h = environment->getScreen().height();
+  size_t screen_size = w*h;
+  
+  assert(grayscale_output_buffer.size() == screen_size);
```
mgbellemare: It's better to not assume that the vector is initialized, and instead allocate the memory using resize(). You can leave the assert() as a sanity check.

mgbellemare:
@mcmachado . Answering your questions: 

1. std::vector will not reallocate the memory if it's already available, so it's better to allocate if needed. See comment.
2. What do you mean by changing the code in two places? Which is the other function?
3. Yes, I think your second fix (row + column, pixel) is better. We certainly should use the same convention across.

mcmachado:
@mgbellemare 

1. I did what I did because it is more efficient. Since you prefer the other way I'll change it.
2. Now we have two functions called `applyPaletteRGB` and two functions called `applyPaletteGrayscale`.
3. In Python we have `[row][column][pixel]`, in C++ it is currently `[row + column][pixel]`. It is not exactly the same (although it is closer). If you prefer I can make it `[row][column][pixel]` but I think `vector<vector<vector> > >` is too convoluted.

When I find time I'll address 1. If you want me to change 3. as well, please let me know.

mcmachado:
@mgbellemare I just changed it. I want to point out the code below, which I implemented, is required. I need to make sure each pixel has allocated 3 positions, but this adds an overhead of the size of the screen, since it has to be iterated twice now.

 ```output_rgb_buffer.resize(screen_size);`
  for(int i = 0; i < output_rgb_buffer.size(); i++){
    output_rgb_buffer[i].resize(3);
  }
```
I made simple tests and everything seems to be working as expected.

**Comment by mcmachado** [comment]
@mgbellemare, do you want me to change anything else on this PR? Or are you happy with it the way it is?

**Comment by mgbellemare** [comment]
Sorry, the earlier notification vanished with holidays. Will take a look and get back to you asap.

File: src/common/ColourPalette.cpp [hunk 3]
```diff
@@ -91,6 +103,15 @@ void ColourPalette::applyPaletteGrayscale(uInt8* dst_buffer, uInt8 *src_buffer,
     }
 }
 
+void ColourPalette::applyPaletteGrayscale(std::vector<unsigned char>& dst_buffer, uInt8 *src_buffer, size_t src_size)
+{
```
mgbellemare: resize() should happen here, not in ALE.

src/common/ColourPalette.cpp
```diff
@@ -81,6 +81,18 @@ void ColourPalette::applyPaletteRGB(uInt8* dst_buffer, uInt8 *src_buffer, size_t
     }
 }
 
+void ColourPalette::applyPaletteRGB(std::vector<std::vector<unsigned char> >& dst_buffer, uInt8 *src_buffer, size_t src_size)
+{
+    uInt8 *p = src_buffer;
```
mgbellemare: resize() should happen here, not in ALE.

mgbellemare:
Okay, two more comments.

1. I didn't realize the Python interface (& other languages) may hand us a raw uInt8 pointer. So that method must stay.

2. In light of (1), I recommend we use a single std::vector<>, rather than a 2-nested vector. This will simplify the code greatly, and make the expected output consistent between the two methods. The role of the std::vector<uInt8> method is then to provide a C++-friendly implementation (i.e. one that does memory management for you).

I can do these changes for you if you're tight for time.

mcmachado:
Marc, I'm tight on time, so if you don't mind I'd rather having you
finishing it. Also, you can make sure the interface is exactly the way you
want.

A reminder: make sure the msnual reflects your changes. You'll just have to
change 2 lines.

Thanks

PR #115, opened 2015-10-15
Adding mode and difficulty selection in the Atari games that support it

mcmachado:
### Summary

This PR adds two features to the ALE: one now is able to select the game mode and the game difficulty to be played. Every result and every experiment reported so far probably used the default mode/difficulty (since changing it was not possible).

In the real Atari console there were several switches (namely, _Power_, _TV Type_, _Left Difficulty_, _Right Difficulty_, _Game Select_ and _Game Reset_ – see figure below). I implemented the _Difficulty_ switches (for one player) and the _Game Select_ switch. 

image

In summary, this opens up several research opportunities. We have many more environments to test our algorithms and we have an easier task to try to address with transfer learning. As a concrete example, below are screenshots of two different game modes in the game `Freeway`.

image

image
### Proper credit

The largest part of this PR was implemented by Nicolas Carion (@alcinos). Robert Post (@Mog333) contributed to the Python interface. I merged the old code to the new version of the ALE, fixing minor issues and warnings. I implemented examples, wrote the documentation and tested the implementation.
### Rationale

One now has access to four new methods: `getAvailableDifficulties`, `getAvailableModes`, `setMode` and `setDifficulty`. They are self-explanatory. 

The mode/difficulty change itself is made mimicking the Stella emulator (see figure below). So, to change a mode we “press” the button _Select_ a specific number of times. Sometimes this has to be done more than once to change a mode. Therefore, we implemented the methods `pressSelect` and `setDifficulty` in `ale_state.cpp` (which in fact are implemented in the `StellaEnvironment`). They are equivalent to the functions implemented in the Stella emulator. We also had to implement the functionality that allows one to write to the RAM (`RomUtils`).

image

Finally, the Stella has a reset button that can be seen as a “soft reset”. We had to implement it as well. Therefore, when one now calls `reset_game` the current mode is also set. Because of that, on the constructor of the classes of emulated games we now set the default variables (lives, score, etc) instead of just calling reset (because a mode would have to be set, which we don’t know yet). The hard reset is equivalent to unplugging the console, which would restore the game to its original state (including the RAM).
### Performed Tests

I ran the examples I added (they use the four introduced functions). My examples work as expected. Our research group has been using this implementation for several months now, we have never had an error, so it should be ok.
### Question

I just added the new available functions to the ALE manual. I did not add a table containing a list of games and their respective number of modes and difficulties. I can do that, but I thought this was too much for the manual, since this is directly related to the supported games. Should I?
### Future(?) improvements

I implemented the methods to be used in the shared interface as well as the Python interface. I did not address the FIFO or the RL-Glue interfaces. 

Also, some games such as Pong have several modes but they are all two-player modes. Because of that, to avoid confusion, I decided to not implement these modes in the ALE. So if one queries Pong he will receive 2 modes because all the others are exclusively two-player modes. Maybe this may be addressed in the future.


mhauskn:
Wow, like the pictures in this PR!

mgbellemare:
@mcmachado I'm starting on this PR now. As you mentioned, it will undoubtedly need updating... if possible please merge #135 first so Continuous Integration can do most of the work.

doc/examples/sharedLibraryInterfaceWithModesExample.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    // Get the vector of legal actions
+    ActionVect legal_actions = ale.getLegalActionSet();
+
+    // Play 10 episodes
+    for (int i = 0; i < modes.size(); i++){
+        for (int j = 0; j < difficulties.size(); j++){
+            
+            ale.setMode(modes[i]);
+            ale.setDifficulty(difficulties[j]);
```
mgbellemare: @alcinos @mcmachado Currently this doesn't reset the environment. The example breaks in Space Invaders when moving from mode 0 to 1. Is this expected behaviour from setMode()/setDifficulty()?

doc/examples/sharedLibraryInterfaceWithModesExample.cpp
```diff
@@ -0,0 +1,80 @@
...
+    ModeVect modes = ale.getAvailableModes();
+    DifficultyVect difficulties = ale.getAvailableDifficulties();
+
+    cout << "Number of available modes: " << modes.size() << endl;
+    cout << "Number of available difficulties: " << difficulties.size() << endl;
+
+    // Get the vector of legal actions
+    ActionVect legal_actions = ale.getLegalActionSet();
+
+    // Play 10 episodes
```
mgbellemare: Minor mismatch between Python and C++ example: this one only plays one episode for each mode/difficulty, the other plays 10 episodes.

src/environment/ale_state.hpp
```diff
@@ -106,6 +116,7 @@ class ALEState {
 
 };
 
-#endif // __ALE_STATE_HPP__
+#include "../games/RomSettings.hpp"
```
mgbellemare: Why is this include there?

src/environment/stella_environment.cpp
```diff
@@ -78,13 +79,15 @@ void StellaEnvironment::reset() {
   emulate(RESET, PLAYER_B_NOOP, m_num_reset_steps);
 
   // reset the rom (after emulating, in case the NOOPs led to reward)
-  m_settings->reset();
+  m_settings->reset(m_osystem->console().system(), *(this));
```
mgbellemare: From a design perspective it would be good to avoid a cyclical dependency between Environment and RomSettings. Was this done because each game needs custom code to switch modes?

src/games/RomSettings.hpp
```diff
@@ -1,92 +1,122 @@
...
+#include <memory>
+#include <stdexcept>
+
+#include "../common/Constants.h"
+#include "../emucore/Serializer.hxx"
+#include "../emucore/Deserializer.hxx"
+#include "../environment/stella_environment.hpp"
+
+class System;
+class RomSettings;
```
mgbellemare: This seems spurious.

src/environment/stella_environment.hpp
```diff
@@ -64,6 +66,18 @@ class StellaEnvironment {
       */
     reward_t act(Action player_a_action, Action player_b_action);
 
+    /** This functions emulates a push on the reset button of the console */
+    void soft_reset();
```
mgbellemare: Should be named 'softReset' for consistency.

src/games/supported/AirRaid.cpp
```diff
@@ -15,8 +15,10 @@
 
 
 AirRaidSettings::AirRaidSettings() {
-
-    reset();
+    m_reward   = 0;
```
mgbellemare: Setting the values here (and in reset() below) breaks the previous semantics of reset(). Do you think we can combine the two back together?

src/games/supported/Alien.cpp
```diff
@@ -31,7 +31,10 @@
 
 AlienSettings::AlienSettings() {
 
-    reset();
+    m_reward   = 0;
+    m_score    = 0;
+    m_terminal = false;
+    m_lives    = 3;
```
mgbellemare: Why do lives not match up between this and `reset(System&,Environment&)`?

src/games/supported/Alien.cpp
```diff
@@ -142,3 +146,42 @@ void AlienSettings::loadState(Deserializer & ser) {
   m_lives = ser.getInt();
 }
 
+//Returns a list of mode that the game can be played in.
+ModeVect AlienSettings::getAvailableModes(){
+    ModeVect modes(4);
+    for(unsigned i = 0; i < 4; i++){
```
mgbellemare: C++11 has a function for this, std::iota. Maybe time to move to C++11?

src/games/supported/Asteroids.cpp
```diff
@@ -131,3 +134,37 @@ void AsteroidsSettings::loadState(Deserializer & ser) {
...
+    modes.push_back(128); //this is the "kids" mode
+    return modes;
+}
+
+//Set the mode of the game. The given mode must be one returned by the previous function. 
+void AsteroidsSettings::setMode(game_mode_t m,System &system, StellaEnvironment& environment){
+    if(m < 32 || m == 128){ /*m >= 0 is implicit, since m is an unsigned int*/
+        m_mode = m;
+        //Read the mode we are currently in
+        unsigned char mode = readRam(&system,0);
```
mgbellemare: Should be reading 0x80 for consistency.

src/games/supported/Atlantis.cpp
```diff
@@ -128,3 +131,29 @@ void AtlantisSettings::loadState(Deserializer & ser) {
   m_lives = ser.getInt();
 }
 
+//Returns a list of mode that the game can be played in.
+ModeVect AtlantisSettings::getAvailableModes(){
+    ModeVect modes(4);
+    for(unsigned i=0;i<4;i++){
```
mgbellemare: Mode 3 is 2-player. Let's make a note of this but disable it here.

src/games/supported/BankHeist.cpp
```diff
@@ -130,3 +133,38 @@ void BankHeistSettings::loadState(Deserializer & ser) {
...
+    }
+    return modes;
+}
+
+//Set the mode of the game. The given mode must be one returned by the previous function. 
+void BankHeistSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+        if(m <= 28 && m % 4==0){ /*m >= 0 is implicit, since m is an unsigned int*/
+        m_mode = m;
+        // Read the mode we are currently in
+        unsigned char mode = readRam(&system,0);
```
mgbellemare: Read from 0x80.

src/games/supported/BankHeist.cpp
```diff
@@ -130,3 +133,38 @@ void BankHeistSettings::loadState(Deserializer & ser) {
...
+        // press select until the correct mode is reached
+        while(mode!=m_mode){
+            environment.pressSelect(1);
+            mode = readRam(&system,0);
```
mgbellemare: This probably works at the beginning, but isn't quite right. The modes are actually defined by bits 2 & 3 of byte 0x80. Bits 0 & 1 encode the maze layout. Let's right shift the byte by 2.

src/games/supported/Berzerk.cpp
```diff
@@ -130,3 +135,42 @@ void BerzerkSettings::loadState(Deserializer & ser) {
   m_lives = ser.getInt();
 }
 
+// returns a list of mode that the game can be played in.
+ModeVect BerzerkSettings::getAvailableModes(){
+    ModeVect modes(9);
+    for(unsigned i = 0; i < 9; i++){
```
mgbellemare: Add comment: this is 1-12 in Atari-decimal (0x01 ... 0x09 0x10 0x11 0x12).

src/games/supported/Berzerk.cpp
```diff
@@ -130,3 +135,42 @@ void BerzerkSettings::loadState(Deserializer & ser) {
...
+// set the mode of the game. The given mode must be one returned by the previous function. 
+void BerzerkSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m >= 1 && (m <= 9 || m == 16 || m == 17 || m == 18)){
+        m_mode = m;
+
+        // we wait that the game is ready to change mode
+        for(unsigned i = 0; i < 20; i++)
+            environment.act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+        // read the mode we are currently in
+        unsigned char mode = readRam(&system, 0);
```
mgbellemare: Read from 0x80 for consistency.

src/games/supported/Boxing.cpp
```diff
@@ -123,3 +125,11 @@ void BoxingSettings::loadState(Deserializer & ser) {
   m_terminal = ser.getBool();
 }
 
+DifficultyVect BoxingSettings::getAvailableDifficulties(){
+    DifficultyVect diff;
+    diff.push_back(0);
```
mgbellemare: Isn't the second bit for the second player? There's only two difficulty choices per player (A and B).

src/games/supported/ChopperCommand.cpp
```diff
@@ -56,13 +60,14 @@ void ChopperCommandSettings::step(const System& system) {
...
     m_terminal = (m_lives == 0);
+    m_isStarted = (readRam(&system, 0xC2) == 1);
 }
 
 
 /* is end of game */
 bool ChopperCommandSettings::isTerminal() const {
 
-    return m_terminal;
+    return (m_isStarted && m_terminal);
```
mgbellemare: Is this a necessary fix to Chopper Command?

src/games/supported/ChopperCommand.cpp
```diff
@@ -129,3 +135,35 @@ void ChopperCommandSettings::loadState(Deserializer & ser) {
   m_lives = ser.getInt();
 }
 
+// returns a list of mode that the game can be played in
+ModeVect ChopperCommandSettings::getAvailableModes(){
+    ModeVect modes;
+    modes.push_back(0);
```
mgbellemare: Add comment: modes 1 and 3 for 2 players.

src/games/supported/CrazyClimber.cpp
```diff
@@ -124,3 +128,37 @@ void CrazyClimberSettings::loadState(Deserializer & ser) {
...
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void CrazyClimberSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m < 4){ /*m >= 0 is implicit, since m is an unsigned int*/
+        m_mode = m;
+        // read the mode we are currently in
+        unsigned char mode = readRam(&system, 0);
```
mgbellemare: 0x80 for consistency.

src/games/supported/CrazyClimber.cpp
```diff
@@ -31,7 +31,10 @@
 
 CrazyClimberSettings::CrazyClimberSettings() {
 
-    reset();
+    m_reward   = 0;
+    m_score    = 0;
+    m_terminal = false;
+    m_lives    = 5;
```
mgbellemare: m_mode not set.

src/games/supported/DemonAttack.cpp
```diff
@@ -59,8 +64,10 @@ void DemonAttackSettings::step(const System& system) {
     // update terminal status
     int lives_displayed = readRam(&system, 0xF2);
     int display_flag = readRam(&system, 0xF1);
-    m_terminal = (lives_displayed == 0) && display_flag == 0xBD;
+    // for terminal checking, we must make sure that we do not detect incorrectly a level change as a game-over
+    m_terminal = (lives_displayed == 0) && display_flag == 0xBD && !m_levelChange;
     m_lives = lives_displayed + 1; // Once we reach terminal, lives() will correctly return 0
+    m_levelChange = false;
```
mgbellemare: Should this variable be called 'm_modeJustChanged' ?

src/games/supported/DemonAttack.cpp
```diff
@@ -121,3 +129,41 @@ void DemonAttackSettings::loadState(Deserializer & ser) {
...
+    modes.push_back(5);
+    modes.push_back(7);
+
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void DemonAttackSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m==1 || m==3 || m==5 || m==7){
```
mgbellemare: Possibly if (m in getAvailableModes()) instead. Actually it could be nice to do this for most games.

src/games/supported/DoubleDunk.cpp
```diff
@@ -116,3 +118,49 @@ ActionVect DoubleDunkSettings::getStartingActions() {
...
+        m_mode = m;
+        //push the select button to open the menu
+        environment.pressSelect(1);
+        //discard the first two entries (irrelevant)
+        environment.act(PLAYER_A_DOWN, PLAYER_B_NOOP);
+        environment.act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+        environment.act(PLAYER_A_DOWN, PLAYER_B_NOOP);
+        environment.act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+        for(unsigned i = 0; i < 4; i++){
+            cout << (m& (1 << i)) << endl;
```
mgbellemare: Do we want to keep cout here? See also using std above.

src/games/supported/DoubleDunk.cpp
```diff
@@ -116,3 +118,49 @@ ActionVect DoubleDunkSettings::getStartingActions() {
...
+// the given mode must be one returned by the previous function
+void DoubleDunkSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m < 16){ /*m >= 0 is implicit, since m is an unsigned int*/
+        m_mode = m;
+        //push the select button to open the menu
+        environment.pressSelect(1);
+        //discard the first two entries (irrelevant)
+        environment.act(PLAYER_A_DOWN, PLAYER_B_NOOP);
+        environment.act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+        environment.act(PLAYER_A_DOWN, PLAYER_B_NOOP);
```
mgbellemare: Why not also make the game length an option? & the player colours? (yes, that's a lot of modes... but lots of opportunities for transfer?)

src/games/supported/Freeway.cpp
```diff
@@ -95,3 +98,37 @@ void FreewaySettings::loadState(Deserializer & ser) {
...
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void FreewaySettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m < 8){ /*m >= 0 is implicit, since m is an unsigned int*/
+        m_mode = m;
+        // read the mode we are currently in
+        unsigned char mode = readRam(&system, 0);
```
mgbellemare: 0x80

src/games/supported/Frostbite.cpp
```diff
@@ -133,3 +137,28 @@ void FrostbiteSettings::loadState(Deserializer & ser) {
...
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void FrostbiteSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m == 0 || m == 2){
+        m_mode = m;
+        // read the mode we are currently in
+        unsigned char mode = readRam(&system, 0);
```
mgbellemare: 0x80

src/games/supported/Gravitar.cpp
```diff
@@ -139,3 +143,48 @@ ActionVect GravitarSettings::getStartingActions() {
...
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void GravitarSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m < 5){ /*m >= 0 is implicit, since m is an unsigned int*/
+        m_mode = m;
+        // read the mode we are currently in
+        unsigned char mode = readRam(&system, 0);
```
mgbellemare: 0x80

src/games/supported/Gravitar.cpp
```diff
@@ -139,3 +143,48 @@ ActionVect GravitarSettings::getStartingActions() {
...
+        // press select until the correct mode is reached
+        while(mode != m_mode){
+            environment.pressSelect(10);
+            mode = readRam(&system, 0);
```
mgbellemare: 0x80

src/games/supported/Pong.cpp
```diff
@@ -96,3 +99,29 @@ void PongSettings::loadState(Deserializer & ser) {
   m_terminal = ser.getBool();
 }
 
+// returns a list of mode that the game can be played in
+ModeVect PongSettings::getAvailableModes(){
```
mgbellemare: There are also four difficulties (two settings for left and two for right).

Also, we've been doing Pong wrong all along. The manual says you can use the fire button to speed up the ball (in both modes). The fire button's behaviour is actually the only difference between the two modes.

src/games/supported/Alien.cpp
alcinos: C++11 could indeed simplify some bits of the code at various places, but was not enabled at the time this PR was written. If the decision is made to move to this standard, I can rewrite the parts were it makes really sense to use new features.
As a side note, some parts of the code use auto_ptr, which has been deprecated. It could be a good opportunity to get rid of them as well

alcinos:
I'm currently updating this PR to incorporate changes in the master branch.
In particular, this means adding the mode/difficulty change for the new games that were added since then.
However, for the game "Donkey Kong", there appear to be two versions of the game (one by Atari and one by Coleco) : [http://www.atariage.com/system_items.html?SystemID=2600&ItemTypeID=ROM&searchValue=donkey](http://www.atariage.com/system_items.html?SystemID=2600&ItemTypeID=ROM&searchValue=donkey)
Which one is the correct one ?

src/games/supported/AirRaid.cpp
alcinos: The problem is that now reset() takes as argument the system and the environment : we need to make the mode persistent to reset, hence we have to call setMode inside reset.
The only way I see of combining them back together is to give system and environment as parameter of the setting constructor (hence we can call reset), but this doesn't seem to be a good idea.
What do you think ?

src/games/supported/DemonAttack.cpp
alcinos: I agree that a pythonic syntax would be nice here. However the c++ equivalent :
` if(std::find(getAvailableModes().begin(), getAvailableModes().end(), m) != getAvailableModes().end()) `
doesn't look that pretty, does it ?

mgbellemare:
@alcinos thanks for looking into this. I believe it's the same game: if you look at the cartridge it has both Atari and Coleco on it.

src/games/supported/DemonAttack.cpp
mgbellemare: With a std::set it's a bit nicer: if (set.count(m) > 0). Still not Pythonic, but close enough?

src/games/supported/StarGunner.cpp
```diff
@@ -146,3 +151,30 @@ void StarGunnerSettings::loadState(Deserializer & ser) {
   m_game_started = ser.getBool();  
 }
 
+// returns a list of mode that the game can be played in
+ModeVect StarGunnerSettings::getAvailableModes(){
```
mgbellemare: Missing difficult switch as well (0-1).

src/games/supported/TimePilot.cpp
```diff
@@ -125,3 +128,10 @@ void TimePilotSettings::loadState(Deserializer & ser) {
   m_lives = ser.getInt();
 }
 
+DifficultyVect TimePilotSettings::getAvailableDifficulties(){
+    DifficultyVect diff;
+    diff.push_back(0);
```
mgbellemare: We need a comment here explaining why there are 3 (and not 4) settings.
(e.g. The manual says the valid settings are 1) L=B, R=B; 2) L=B, R=A; 3) L=A, R=B, with 3) being the hardest.)

doc/examples/sharedLibraryInterfaceWithModesExample.cpp
mgbellemare: Following up on the question of setDifficulty() resetting: a number of games require a reset after setting the difficulty. We should do so. Is there a reason why difficulty is stored in StellaEnvironment, but mode is stored in ROMSettings? Would it make more sense to store both in ALEState and serialize them?

src/games/supported/VideoPinball.cpp
```diff
@@ -125,3 +129,36 @@ void VideoPinballSettings::loadState(Deserializer & ser) {
...
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void VideoPinballSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m == 0 || m == 2){
+        m_mode = m;
+        //Read the mode we are currently in
+        unsigned char mode = readRam(&system, 0xC1);
```
mgbellemare: 0xA1 seems to be the correct byte here. 0xC1 is the number of select presses since the last reset. In particular setMode(0) will keep the mode the same (even if it is mode 2). Can you check?

src/games/supported/YarsRevenge.cpp
```diff
@@ -134,3 +138,41 @@ ActionVect YarsRevengeSettings::getStartingActions() {
...
+    modes.push_back(0x20);
+    modes.push_back(0x40);
+    modes.push_back(0x60);
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void YarsRevengeSettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment){
+    if(m == 0 || m == 0x20 || m == 0x40 || m == 0x60 ){
```
mgbellemare: The manual states some differences between mode 0 and 0x20, but I can't see these visually. Do you remember what the differences are?

src/games/supported/Zaxxon.cpp.orig
```diff
@@ -0,0 +1,170 @@
+/* *****************************************************************************
```
mgbellemare: Is this file still necessary?

src/games/supported/YarsRevenge.cpp
alcinos: The difference is the speed of the auto guided missile : in mode 0, it takes 12seconds to reach the glittering area, vs only 10 seconds in mode 0x20. (yes the difference is tiny)
However, I noticed that the default mode of this game is 0x20, I'll init m_mode accordingly

src/games/supported/Zaxxon.cpp.orig
alcinos: Git Merge artefact. Removing it.

src/games/supported/DemonAttack.cpp
alcinos: Agreed, we loose the ordering of the modes, but it doesn't make much sense in the majority of the game anyways.
If you agree to c++11, we can then have a very nice syntax for the getAvailableModes, using the {} constructor:
`   return {1,3,5,7}`, which looks nicer than a bunch of inserts/push_back.

doc/examples/sharedLibraryInterfaceWithModesExample.cpp
alcinos: I'll look into the issues of resetting.
For the question of storage of difficutly and mode params, the reason was that the difficulty is something console-related (the physical position of the switches) while the mode is game-specific (which means custom change code).
Because of that, if we want to have both parameters stored in the same class, IMO it would make more sense in ROMSettings. But in any case, it would be weird somehow to have the parameter stored at one place, while the code that actually performs the modification is somewhere else.

About the serialization : do we want anyway to serialize these parameters ? I'm not sure to see the purpose.

src/games/supported/DemonAttack.cpp
mgbellemare: All right, I think we should move to C++11. It's overdue now and there's a number of features here that will be much nicer in C++11. @mhauskn , @mcmachado , can you think of a reason not to move to C++11?

It probably makes sense to have a separate PR first moving ALE 0.5.1 to C++11 (in the dev branch), then continue the work here. WDYT?

doc/examples/sharedLibraryInterfaceWithModesExample.cpp
mgbellemare: I agree the way it currently is aligns with the console. From an RL perspective though setMode & setDifficulty should have a consistent effect across games. I think the easiest is to force a reset, since most games require it.

Serialization needs to restore the game state to what it was, including the current difficulty. E.g. if you're performing tree search, you don't want to have to remember what difficulty setting was on (again this is more
an RL interface vs Atari emulation question).

I'm okay with the parameters being stored in either ALEState or ROMSettings -- whichever you think is most relevant.

src/games/supported/DemonAttack.cpp
mcmachado: @mgbellemare I agree, we should move to C++11 and I can't think of a reason not to. However, as you said, we should do it in a different PR, I think.

mhauskn: OSX doesn't support c++11's Thread Local Storage (TLS). The status of
support for various c++11 features can be seen at
http://clang.llvm.org/cxx_status.html

On Tue, Mar 29, 2016 at 5:00 PM, Marlos C. Machado <notifications@github.com

alcinos: @mhauskn I don't really see your point since we don't use (and AFAICT don't plan to) use c++11 threading capabilities

mhauskn: My point is only to be careful with the c++11 features we do decide to use,
as support can be spotty on OSX.

mgbellemare: Fair enough. My recommendation is to keep the C++11 to the less-exotic operations. I'll try to get a C++11 PR out this weekend.

nczempin:
Is this just about some minor conflicts that can quickly be resolved, or are there still too many unanswered review comments?

I'd like this feature in general, but I would prefer to get an as-simple-as-possible version, possibly only supporting one game, with the others falling back to the current behaviour as default.

I'm thinking about how to add 2-player games with the agent playing both sides ("self-play"), and the general idea of this PR seems to be a stepping stone.

I'm also hoping that this approach wouldn't contradict severely with adding a game like Space Shuttle, although if that's the only game that uses the switches in creative ways I'd accept that it's not worth making major changes just to support it.

mgbellemare:
I believe the main hurdle came from a needed conversion to C++11. I believe there are also unresolved comments. This is certainly a much needed PR. If you want to help out clean it up so we can bring it into ALE 0.6, that'd be great.

nczempin:
I'll have a look at it. Like I said, I would prefer to provide minimally-invasive PRs, so for example I don't see how a C++11 conversion is strictly _needed_ for the basic functionality to work.

What's the current status on C++11? I saw an unmerged PR from yourself, are you holding this back for some reason? Would changing to 11 cause breakages for some people? I have seen some inconsistencies in C++11 between MacOS, Windows and Linux, but they were in areas unlikely to be necessary for ALE.

nczempin:
I'll check if I can just resolve the conflicts and check the review comments, but if it seems too complicated I'll probably just do a rewrite that's a bit more incremental.

mgbellemare:
If you're needing the 2-player code urgently, it might be easier to patch in what you need. Then again, it would be great to get this PR sorted out @mcmachado

nczempin:
> If you're needing the 2-player code urgently, it might be easier to patch in what you need.

I don't really need any of this, urgently or not. It would just be more useful in general to take incremental steps in case the entirety of this PR turns out to be involved. I haven't really looked at it yet, beyond sorta looking through the comments and noticing there's a conflict (which may well just be due to the PR's age and could be sorted or at least simplified with a simple rebase.

I'm hoping I can find some time during the coming week at least to see what the situation is.

src/environment/ale_state.hpp
nczempin: I put it in the .cpp

src/games/RomSettings.hpp
nczempin: done in my PR

src/environment/stella_environment.hpp
nczempin: done in my PR

src/games/supported/BankHeist.cpp
nczempin: 0 is converted to 0x80 in readRam. IIRC there are several places in which the offset is used vs. where it is not used.

src/games/supported/Berzerk.cpp
nczempin: I changed the 16 17 18 to 0x10 0x11 0x12. Not sure the comment is still needed, but I put it in.

mcmachado:
@mgbellemare @nczempin: What is the current status of this PR?
I submitted it a long time ago and now I have some time to try to be done with it. I do realize there is also [PR 196](https://github.com/mgbellemare/Arcade-Learning-Environment/pull/196) now that is working on the same things.

How can we solve all this?

nczempin:
Well, you can migrate the changes I made in PR 196 (to formally resolve the conflicts e. g. with the manual) over here and then you could address all the open review comments here.

I have started working on a simplified version of the issues this PR is trying to solve, but I won't mind if you simply get this one working.

mcmachado:
@mgbellemare what do you prefer?

@nczempin what did you simplify from this PR?

nczempin:
I simplified based on the idea that it is not necessary to wait until _all_ roms support _all_ their modes to get useful functionality out of this change.

Many of the review comments seem to be isolated in individual rom support files that have no bearing on the other roms, with some of the comments asking for more info and some perhaps requiring a more detailed discussion.

On the other hand just getting a single rom with support for modes would be useful, and that could also be thoroughly tested.

Another issue may be that this PR may be so old that the parent branch may have moved on considerably in the meantime.

So I was basically just going for a "minimum viable product" that's based of the _develop_ branch (so that C++11 support can be assumed, which is also not strictly necessary, but more convenient).

However, it is difficult to predict how much time I'll be allocate to such activities; for me there is the additional work of having to understand someone else's code, whereas for you it's more along the lines of remembering your own code.

mcmachado:
@nczempin This seems to be the right thing to do (to not depend on all ROMs being updated).
Regarding "remembering my own code", I wouldn't take that into consideration. It was two years ago and @alcinos wrote a big part of it.

nczempin:
Well, you mentioned that you have some time now to deal with this, so do
feel free to take over; I haven't really got very far.

If you want I can dig out what I did to get started on this issue, although
it may not be very much.

mgbellemare:
I also have relatively limited time to sort this out, but I'll have a look again this weekend. The main issue with the original PR is that it was monolithic. I believe @nczempin 's PR was much easier to deal with in that respect. So that would be my preferred way going forward.

mcmachado:
Should we close this PR then, to focus on the other one?

mcmachado:
This feature was implemented in a couple of smaller PRs recently. I'm closing it now.

PR #132, opened 2016-01-03
Add support for the game Lost Luggage.

nowozin:
The ROM file is `lostluggage.bin`, with the following information:

```
  Cart Name:       Lost Luggage (1981) (Apollo)
  Cart MD5:        7c00e7a205d3fda98eb20da7c9c50a55
  Controller 0:    Joystick in left port
  Controller 1:    Joystick in right port
  Display Format:  NTSC*
  Bankswitch Type: 4K* (4K) 
```

The lives are handled slightly non-standard in this game: with one life (luggage icon in the bottom left) remaining the game ends if you miss a luggage piece.  Is there any guidance on how this should be handled?  (I can subtract 1 of course but the current commit just outputs the in-game display of lives.)

Lost Luggage screenshot

### DIFF
--- a/src/games/supported/LostLuggage.hpp
+++ b/src/games/supported/LostLuggage.hpp
@@ -1,0 +1,78 @@
+/* *****************************************************************************
+ * The lines 70 and 77 are based on Xitari's code, from Google Inc.
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License version 2
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ * *****************************************************************************
+ * A.L.E (Arcade Learning Environment)
+ * Copyright (c) 2009-2013 by Yavar Naddaf, Joel Veness, Marc G. Bellemare and 
+ *   the Reinforcement Learning and Artificial Intelligence Laboratory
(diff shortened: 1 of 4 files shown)


src/games/supported/LostLuggage.hpp
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * *****************************************************************************
+ */
+#ifndef __LOSTLUGGAGE_HPP__
+#define __LOSTLUGGAGE_HPP__
+
+#include "../RomSettings.hpp"
+
+/* RL wrapper for Lost Luggage */
+class LostLuggageSettings : public RomSettings {
```
mgbellemare: 'Fire' resets the game, so you should also implement isLegal for this game. See Skiing.{hpp,cpp} for details.

mgbellemare:
I'm not sure I follow the lives comment -- is the issue that in many Atari 2600 games, no lives displayed == 1 life left, and this follows a different convention?

mgbellemare:
The convention I've followed for lives is: it should reach 0 at the same time as isTerminal() becomes true (unless it's a game with no lives, in which I just output 0). It seems like you did the correct thing here, no?

nowozin:
@mgbellemare Thanks; I will fix `isLegal` and the definition if lives tomorrow, right now lives are matched with the displayed number of lives, not the actual number of lives.

mgbellemare:
Aren't there 3 lives, though? So that the number of pieces of luggage at the bottom correctly describes the number of lives?

Thanks for doing all this great work, it's very refreshing to see support for some more Atari 2600 games! One comment: if you add any more games, please put them all in the same PR. It'll make things a lot simpler on my end.

nowozin:
@mgbellemare I double checked, lives were indeed handled correctly.  Also I added isLegal to prevent any FIRE action, as is done in `Skiing.cpp`.

nowozin:
@mgbellemare Just to understand your workflow: why would multiple games in one PR be easier to handle?  (I just checkout the branch, do a make, and that is quite convenient; is there a better way?)

mgbellemare:
Individual PRs involve a lot more manual work. For example this PR needs to be resolved manually now, because git diff is failing to understand the changes. Also it would have allowed me to test all the new games without needing to pull 10 different branches, switch between them, etc.

PR #137, opened 2016-01-10
Fix terminal condition.

nowozin:
This fixes the wrong terminal condition.

I noticed however, that sometimes the `sharedLibraryInterfaceExample` application hangs on special events such as playing in-game sounds and game-over.


mgbellemare:
Yes, I've noticed this as well. This occurs when in manual control mode.

mgbellemare:
LGTM. What a hard game!

PR #142, opened 2016-01-29
Quick and dirty compilation instructions

alito:
I compile ALE infrequently enough that I always forget which method it uses and forget where to look. The twin compilation systems don't help.

I think adding this instructions to Readme.txt could help other people too. Might be worth adding them in README.md too, but that one points to the manual already. The manual is a little bit too much for people who just want to compile and play though, so it might still be worth it. 

Thoughts?

### DIFF
--- a/Readme.txt
+++ b/Readme.txt
@@ -15,6 +15,23 @@ Enjoy,
 The ALE team
 
 ===============================
+Quick start
+===============================
+
+Compilation:
+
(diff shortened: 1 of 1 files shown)


mgbellemare:
Very good idea. Do you think it makes sense to put this under README.md instead, so that it's directly visible on the GitHub landing page?

alito:
I think it does, but it's a matter of taste.
I've created a separate pull request that also adds it to README.md. You can see what it looks like here:
https://github.com/alito/Arcade-Learning-Environment/tree/compilationinstructions2

PR #143, opened 2016-01-30
Add support for Galaxian

alito:


mgbellemare:
LGTM. Thanks!

src/games/supported/Galaxian.cpp
```diff
@@ -0,0 +1,130 @@
+/* *****************************************************************************
+ * The lines 54 - 59, 100, 109 and 117 are based on Xitari's code, from Google Inc.
```
mgbellemare: Can you take out this comment, since it isn't true for this particular game?

alito:
Gah, sorry. I templated off SpaceInvaders and I didn't notice that line.

PR #145, opened 2016-02-02
Updated ale_python_interface for python3.

mhauskn:
Minimal updates required for python2/3 compatibility. I'm unhappy with having to append 'b' in front of strings when interacting with ale (e.g. ale.setInt(b'random_seed', 123)). Reason is because strings in python3 are unicode by default while ctypes expects a byte-encoded string (which used to be default in python 2.7). Adding the b prefix tells python3 to treat the string as bytes. Hopefully ctypes will be updated to play better with python3 unicode strings. Until then, I'm afraid we may be stuck with the b-prefixed strings. 

I've tested the PR on Ubuntu with python 2 and python 3. Both work fine, and pip install is working for both as well.

### DIFF
--- a/ale_python_interface/__init__.py
+++ b/ale_python_interface/__init__.py
@@ -1,1 +1,1 @@
-from ale_python_interface import *
+from .ale_python_interface import *
--- a/doc/examples/python_example.py
+++ b/doc/examples/python_example.py
@@ -9,13 +9,13 @@ from random import randrange
 from ale_python_interface import ALEInterface
 
 if len(sys.argv) < 2:
-  print 'Usage:', sys.argv[0], 'rom_file'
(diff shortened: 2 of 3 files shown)


ale_python_interface/__init__.py
```diff
@@ -1 +1 @@
-from ale_python_interface import *
+from .ale_python_interface import *
```
mgbellemare: Is the period there on purpose?

mhauskn:
Yes, it was recommended by '2to3', python automatic python 2 --> python 3 conversion program. I think (but haven't verified) that the period may be responsible for not having to do odd import statements, such as the ones described in the thread:

from ale_python_interface.ale_python_interface import *

PR #146, opened 2016-02-03
Added fifo example to CMakeLists; fixed a read bug in FIFO controller

mgbellemare:
A minor fix to the FIFO controller, which was causing a crash in the fifo example. Please LGTM @mhauskn or @mcmachado


mhauskn:
This looks good to me, assuming the fifo process will terminate on it's own after fgets returns null. Also, there is one more unsafe call to fgets on line 101.

PR #147, opened 2016-02-11
Generating pkg-config file

pierrelux:
This pull requests adds the ability to generate a [pkg-config](https://people.freedesktop.org/~dbn/pkg-config-guide.html) file. Pkg-config simplifies the compilation process by providing the user with the ability to obtain the correct cflags and ldflags. In my case, this results in:

```
$ pkg-config --cflags --libs libale
 -I/home/pierre-luc/workspace/Arcade-Learning-Environment/src  -L/home/pierre- luc/workspace/Arcade-   Learning-Environment -lale -lz -lm
```

Running `cmake CMakelists.txt` with the following changes will automatically create a `libale.pc` from the `libale.pc.in` template. `.pc` files are usually installed somewhere under `/usr/lib/pkgconfig` but since the the current `CMakeLists.txt` lacks install targets, `libale.pc` will simply stay at the root of the source directory. 

In this case, the user can still use the `.pc` file by setting the `PKG_CONFIG_PATH` environment variable. In my case, something like: 

```
$ export PKG_CONFIG_PATH=$HOME/workspace/Arcade-Learning-Environment
```

or upon calling `make` or `gcc`:

```
$ PKG_CONFIG_PATH=$HOME/workspace/Arcade-Learning-Environment make 
```

where `make` here reads a Makefile which uses `pkg-config`. 

On line 4 of `CMakeLists.txt`, I added a variable `ALEVERSION` which is then used in the `.pc.in` template. Note that ideally, we could extract the version number automatically from the `ChangeLog` using a `STRING (REGEX MATCH ...)` cmake command. I try to write this but had no luck to get cmake to understand my regex. Note that just making sure to maintain version numbers in synced between `CMakeLists.txt` and `ChangeLog` would also be a valid solution. It just requires a bit more human involvement. In any cases, unless the user really structure his or her compilation pipeline in such a way that it depends heavily on `pkg-config --modversion`, having an out-of-synced version number should have no consequence. 

I recently wrote a Python wrapper for @mcmachado "shallow" features. Having such a pkg-config feature would make it easier not to hardcode absolutes paths into the `setup.py`. 

The proposed change is expected to work on all platforms, with or without pkg-config. The `.in` business purely depends on `cmake` and does not require `pkg-config` to be installed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required (VERSION 2.6)
 project(ale)
 
+set(ALEVERSION "0.5dev_b")
+set(PACKAGE_NAME "Arcade Learning Environment")
+
 option(USE_SDL "Use SDL" OFF)
 option(USE_RLGLUE "Use RL-Glue" OFF)
 option(BUILD_EXAMPLES "Build Example Agents" ON)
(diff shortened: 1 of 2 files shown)


marcgb:
Thanks for the numerous updates, @pierrelux . I'll go over them as soon as I find a bit of time.

mgbellemare:
LGTM. The current tagged version is 0.5.1, but master is slightly ahead of that. I'm not sure how to properly deal with changing versions -- hardcoding it into CMake introduces a dependency which is almost surely going to be broken. You suggest that pkg-config is not too sensitive to versions. As some kind of compromise, maybe we can set the version to 0.5 and only change it when we move to the next minor version?

JesseFarebro:
Hi @pierrelux,

I think this can now be closed. Since you originally posted this PR I overhauled the build system quite a bit.
I was debating including a pkg-config during the rewrite but decided on just installing the CMake package config instead. With C++20 modules coming this will probably change further. Thanks!

PR #153, opened 2016-02-24
Installing the C++ targets for Unix

jeremyfix:
These modifications of the main CMakeLists.txt installs the targets (binaries, libraries and headers)

I'm not sure of where to install these for Apple and Windows so the install commands are activated only for Unix.

Therefore to install , under Unix : 

cmake -DUSE_SDL=ON -DUSE_RLGLUE=OFF -DBUILD_EXAMPLES=ON -DCMAKE_INSTALL_PREFIX=/usr
make -j 4
make install

This installs the binaries in   /usr/bin , adding the prefix "ale-" to the binaries
                    the headers in /usr/include/ale/
                    the libraries in /usr/lib

If you wish the libraires to be installed in   /usr/lib64 , you simply need to modify the cmake call :
cmake -DUSE_SDL=ON -DUSE_RLGLUE=OFF -DBUILD_EXAMPLES=ON -DCMAKE_INSTALL_PREFIX=/usr -DLIB_SUFFIX=64
make -j 4
make install


pierrelux:
If this PR is accepted, I should update https://github.com/mgbellemare/Arcade-Learning-Environment/pull/147 accordingly (which remains to be accepted as well).

jeremyfix:
I added the generation of a pkg-config file. Rather than configuration a file, the pkg-config file is directly generated from the CMakeLists.  I did this only for UNIX systems.

mgbellemare:
@jeremyfix Thanks for the PR, this brings us a step closer to having a more standardized ALE install. Out of curiosity: was this done because you found it otherwise cumbersome to link your projects to the ALE?
@pierrelux Since this PR is a bit larger I'll merge it first and let you update #147 accordingly.

CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required (VERSION 2.6)
 project(ale)
+set(ALEVERSION "0.5dev_b")
```
mgbellemare: The version question should be addressed now if possible, since this isn't too practical. See #147.

CMakeLists.txt
```diff
@@ -74,12 +125,26 @@ include_directories(
...
+if(UNIX)
+	install(TARGETS ale-lib 
+		DESTINATION ${LIB_INSTALL_DIR})
+endif()
+
 add_executable(ale-bin ${SOURCE_DIR}/main.cpp ${SOURCE_DIR}/ale_interface.cpp ${SOURCES})
 set_target_properties(ale-bin PROPERTIES OUTPUT_NAME ale)
 set_target_properties(ale-bin PROPERTIES RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
+if(UNIX)
+	install(TARGETS ale-bin
```
mgbellemare: Do you think the examples need to be installed? I'd prefer to keep the ALE's install footprint as small as possible.

mgbellemare:
With your changes, ALE's `cmake` under Mac OS X also supports `make install`. On my machine it defaults to /usr/local, which seems appropriate. However, there doesn't seem to be a `make uninstall` command. Is it possible to add `make uninstall` functionality?

CMakeLists.txt
jeremyfix: Ok, I will change it to 0.5

CMakeLists.txt
jeremyfix: I like to install the examples as well . But we are not obliged to. I will drop them out from the install.

jeremyfix:
If you wish to change the installation prefix you just need to provide -DCMAKE_INSTALL_PREFIX=/my/path  when you invoke cmake , e.g.
cmake -DCMAKE_INSTALL_PREFIX=/usr ....  <- will install all the things in /usr/bin , /usr/lib, ..
cmake ....  <- will by default install in /usr/local/bin, /usr/local/lib, ... on Unix systems as well

I will add an uninstall target following this recipe : https://cmake.org/Wiki/RecipeAddUninstallTarget

mgbellemare:
`make uninstall` works great, thanks. Was there a reason for installing all the header files? I can see that ale_interface.hpp depends on other includes right now. It would be good to reduce this dependency to a minimum, e.g. a single ale_interface.hpp (but this is perhaps beyond the scope of this PR).

jeremyfix:
I do not get your point for the headers; We can set a master header ale.hpp which includes all the required headers so that developers just have to include ale.hpp which itself includes all the required files. But we necessarily have to install them all as they are all required; otherwise, that would mean that part of the code is not required; 

As a side note, I would suggest to remove the makefile.*  at the root dir as the compilation is handled with cmake to avoid confusion on how to build the project;

mgbellemare:
Hi @jeremyfix , sorry for the delay in getting back to you about this (reviewing season happened!). In the best of worlds C++/Python users shouldn't need to interface with anything else than ale_interface.hpp
and possibly ALEState.hpp. The extra headers are only there because of member variable dependencies, but it's possible to get rid of these via the [pImpl design pattern](https://tonka2013.wordpress.com/2013/08/31/why-every-c-developer-should-know-about-the-pimpl-idiom-pattern/).

Yes, the makefile.* are getting old now. They were kept there so that the installation process would be simple for newcomers to the ALE. But I agree with you it is starting to get in the way.

jeremyfix:
That was intended; 
I find it easier to browse the examples provided by the project by typing : ale<tab><tab> and benefit from completion to get the list of the binaries.

I also like to prefix the binaries by the project name because that is a unique prefix for all the provided binaries . They do not interfere with binaries provided by someone else that might have the same name.

mgbellemare:
Right, when the examples are installed it makes sense. It seems less necessary with if they aren't being installed, but I can see the namespace-like benefits of the approach.

PR #159, opened 2016-04-15
Avoid build noise in versioning

shelhamer:
In ALE master compiling the project generates a number of outputs, like the CMake build files and the `ale` binary itself, that can interfere with hacking on ALE and versioning changes.

For my work I decided to git ignore these generated files and use a build dir to keep the CMake files. This way I don't accidentally commit these generated outputs and I dont see the dirty status in my projects that include ALE. I thought I'd send a PR in case you think this is more convenient too.

Thanks for ALE!

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ The Arcade Learning Environment (ALE) -- a platform for AI research.
 ### To compile:
 
 ```
-$ cmake -DUSE_SDL=ON -DUSE_RLGLUE=OFF -DBUILD_EXAMPLES=ON .
+$ mkdir build && cd build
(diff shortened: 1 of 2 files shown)


mgbellemare:
LGTM, thank you!

README.md
```diff
@@ -11,7 +11,8 @@ The Arcade Learning Environment (ALE) -- a platform for AI research.
 ### To compile:
 
 ```
-$ cmake -DUSE_SDL=ON -DUSE_RLGLUE=OFF -DBUILD_EXAMPLES=ON .
+$ mkdir build && cd build
```
mgbellemare: Not everyone uses a `build` directory, but I agree it makes for much cleaner output.

PR #160, opened 2016-04-17
Onwards to c++11

mgbellemare:
This makes the minimum changes to the ALE to make it compile with C++11 features. The continuous integration build is passing, which gives me some reassurance that this will mostly work. @mcmachado @mhauskn @alcinos please compile this PR on your machines and let me know if something needs addressing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ install:
 
 addons:
     apt:
+        sources:
+            - ubuntu-toolchain-r-test
         packages:
             - g++-4.8
             - clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,21 @@ language: cpp
 #install:
 #    - sudo apt-get install libsdl1.2-dev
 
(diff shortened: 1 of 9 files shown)


mgbellemare:
There is also the related question of whether this should be brought into the dev branch first, and further modifications made there.

mcmachado:
@mgbellemare I just downloaded and compiled this PR. Aside from CMake importing the wrong SDL library, which always happens to me, everything worked fine. I compiled with and without SDL. I compiled my research code linking with the new ALE and everything also worked fine. Again, I tested with and without screen display.

I didn't test RL-Glue because I don't have any code which I can use, but I'm ok with you merging this PR.

mcmachado:
Oh! And regarding the dev branch, I think it is probably best to merge this PR to it first. We ensure we'll have more time to test since we will be forced to use c++11 in all the other PRs we are working on.

mgbellemare:
Okay, merging this into the dev branch.

PR #161, opened 2016-04-21
Fixed getScreenRGB, was not sending all three channels.

thelukester92:
getScreenRGB was only sending the first third of pixels to the buffer because it did not multiply the screen size by the number of channels (3 in case of RGB). Fixed.

### DIFF
--- a/src/ale_interface.cpp
+++ b/src/ale_interface.cpp
@@ -259,7 +259,7 @@ void ALEInterface::getScreenGrayscale(std::vector<unsigned char>& grayscale_outp
   size_t screen_size = w*h;
   
   pixel_t *ale_screen_data = environment->getScreen().getArray();
-  theOSystem->colourPalette().applyPaletteGrayscale(grayscale_output_buffer, ale_screen_data, screen_size);
+  theOSystem->colourPalette().applyPaletteGrayscale(grayscale_output_buffer, ale_screen_data, screen_size*3);
 }
 
 //This method should receive a vector to fill it with
@@ -272,7 +272,7 @@ void ALEInterface::getScreenRGB(std::vector<unsigned char>& output_rgb_buffer){
 
   pixel_t *ale_screen_data = environment->getScreen().getArray();
 
-  theOSystem->colourPalette().applyPaletteRGB(output_rgb_buffer, ale_screen_data, screen_size);
+  theOSystem->colourPalette().applyPaletteRGB(output_rgb_buffer, ale_screen_data, screen_size*3);
 }
 
 // Returns the current RAM content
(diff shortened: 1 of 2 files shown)


mgbellemare:
Do you have code illustrating this issue? Looking at the code everything seems fine -- src_size in applyPaletteRGB() should be the number of pixels in the source buffer, i.e. w * h.

thelukester92:
Given a savePng function that uses libpng, here's the code that causes a problem:

	ALEInterface ale;
	ale.loadROM("roms/Breakout.bin");
	
	size_t width = ale.getScreen().width();
	size_t height = ale.getScreen().height();
	size_t channels = 3;
	
	vector<pixel_t> buffer(width * height * channels);
	ale.getScreenRGB(buffer);
	
	png_bytep *byteData = new png_bytep[height];
	for(size_t i = 0; i < height; i++)
	{
		byteData[i] = new png_byte[width * channels];
		for(size_t j = 0; j < width; j++)
			for(size_t k = 0; k < channels; k++)
				byteData[i][channels * j + k] = (png_byte) buffer[i * width * channels + j * channels + k];
	}
	
	savePng("test.png", byteData, width, height, channels);
	return 0;

Attached are images illustrating the problem. The first is what happens with your code. Only the first third of the pixels are filled because getScreenRGB only gets width x height values, rather than width x height x channels (thus only getting the first ~30k values instead of the full ~100k values needed for RGB). The second is what happens with my fix, by multiplying by the number of channels, we get the full ~100k values and thus see the whole screen.

test-bad
test-good

jleni:
I agree that it doesn't work properly in this version. It seems that a bug got introduced in d5b1c54 when an additional function using std::vectors was introduced.

@thelukester92: The arguments used in the calls are correct, I would not apply these changes. Instead I would suggest that the correct fix is to change `ColourPalette.cpp` `line 91`:

From:     
`for(size_t i = 0; i < src_size; i += 3, p++){`
To:         
 `for(size_t i = 0; i < src_size*3; i += 3, p++){`    // Notice the *3 there

mgbellemare:
Oops, my bad. @thelukester92 , @jleni 's fix is the right one. Do you want to revert your changes and add this?

thelukester92:
Fix applied, now using @jleni's fix.

mgbellemare:
Great. Thanks for the fix!

ppwwyyxx:
The above fix seems wrong. `screen_size * 3` was passed to ColourPalette, and then ColourPalette also has `src_size * 3`. This ends up being `screen_size * 9`.
Found this because `valgrind` is complaining about memory error inside ALE:
```
==3901374== Invalid read of size 1
==3901374==    at 0x4FD00B3: ColourPalette::applyPaletteRGB(std::vector<unsigned char, std::allocator<unsigned char> >&, unsigned char*, unsigned long) (in /home/xxx/.local/lib/libale.so)
```
Removing `*3` in `ale_interface.cpp` helps.

PR #167, opened 2016-05-15
Make StellaEnvironment::isTerminal and ALEInterface::game_over const

rhaps0dy:
It came up when I was using ALEInterface in an old MDP class. Not very useful but it costs little effort.

Also removed unnecessary duplicate checks on ALEInterface::game_over


mgbellemare:
LGTM, thanks!

PR #168, opened 2016-05-21
Default color averaging to false, making observations single frames

shelhamer:
This turns off color averaging by default.

By default, and counter to the documentation, ALE blends frames together to make environment observations. I would like to argue that it is better to define an observation as a single frame by default as it more closely corresponds to the "true" emulator screen. This is the approach taken by deepmind/xitari for instance, and it might be nice to align settings in this case to help avoid fragmentation.

I was burned by this misdocumented default, rendering certain experiments less than perfectly comparable, so I'm sending this PR in case it is accepted and helps others avoid this issue.

I included a slight doc change and a changelog notice in case these are helpful and desired, but I could leave these out too if you prefer.

This is an opinionated PR, but it is made with the best intention. Thanks for the ALE!

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+Inter-release notes:
+  * color_averaging is now off by default so that environment observations correspond to emulator frames unless requested otherwise.
+
 October 4th, 2015. ALE 0.5dev_b.
   * Enforce flags existence (@mcmachado).
   * Fix RNG issues introduced in 0.5.0.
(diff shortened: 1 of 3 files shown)


mgbellemare:
LGTM. This is a genuine bug: color averaging should be off by default -- my apologies for the wasted science time.

Do be aware that almost all DQN-style results use luminance max-pooling, which is quite similar.

@mcmachado I think the setting was accidentally set to true when we moved the defaults to Settings.cxx. We should implement a way of ensuring defaults don't change between commits.

mcmachado:
@mgbellemare we had a lot of back and forth on this, but yes, we should have a way to make sure the defaults are never changed accidentally.

PR #169, opened 2016-05-21
Default to deterministic action (set action repeat stochasticity to 0)

shelhamer:
This PR makes emulator action deterministic by default. 

I know that this setting was decided at the AAAI-15 workshop and that it discourages "degenerate" policies for certain games. However, I argue that lurking nondeterminism leads to confusion and frustration. While I appreciate the functionality of this option, and the manual's explanation of it, it is a dangerous default that can invalidate experiments and recorded data. Furthermore, when `repeat_action_probability > 0` it is impossible to know which action was in fact taken in the emulator. This is not exposed through the interface (at least as far as I can tell), making it impossible to record true trajectories of agents or humans. For these reasons a default of 0 is the safest, and those concerned with robustness above all else can request it explicitly.

This mirrors the setting in deepmind/xitari to hopefully reduce fragmentation. As the ALE grows in popularity, it could be helpful to unify work on it.

With #168, this better aligns the ALE defaults to the expectation that the agent is interacting with the emulator environment as-is for input and output.

This is an opinionated PR, but it is made with the best intention. Thanks for the ALE!

@rhaps0dy I believe you have also had trouble with this setting.

### DIFF
--- a/doc/manual/manual.tex
+++ b/doc/manual/manual.tex
@@ -432,18 +432,16 @@ command--line argument \verb+-color_averaging+ (or the \verb+setBool+ function).
 
 \subsection{Action Repeat Stochasticity}
 
-Beginning with ALE 0.5.0, there is now an option (enabled by default) to add 
-\emph{action repeat stochasticity} to the environment. With probability $p$ (default: $p = 0.25$),
-the previously executed action is executed again during the next frame, ignoring the agent's
-actual choice. This value can be modified using the option \verb+action_repeat_probability+.
-The default value was chosen as the highest value for which human play-testers
-were unable to detect any delay or control lag.
+Beginning with ALE 0.5.0, there is now an option (disabled by default) to add \emph{action repeat stochasticity} to the environment.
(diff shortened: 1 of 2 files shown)


mgbellemare:
Hi @shelhamer ,

I agree with you that "lurking nondeterminism" is bad. Actually, the Atari 2600/Stella itself is not completely deterministic: it depends on the emulator's start conditions, and software resets actually don't result in the same start state. I spent a good deal of time exactly writing the ALE code to make things deterministic, so it's a bit ironic now that I'm advocating for the opposite. But from an RL research perspective I really believe we should be transitioning away from the deterministic environment.

What I'm hearing from you, however, is that there's a disconnect between the ALE-as-a-blackbox and the ALE-as-a-sandbox. It would be great to have proper tools to handle recording of trajectories, etc., rather than individuals having to re-implement the basics. As you point out, the ALE doesn't expose the actions taken (this is so that the environment is properly stochastic from the agent's perspective). Why not add some code to ALEInterface to separately record the internal dynamics, in particular the trajectories, and also replay it?

PR #173, opened 2016-06-29
Added code necessary to support Surround

mgbellemare:
This is adding support for Surround. This requires adding the select button, leading to a little bit of redundancy between the modes PR and this one. When the modes PR is complete, we'll need to remove the starting actions from this game's settings.

### DIFF
--- a/src/games/Roms.cpp
+++ b/src/games/Roms.cpp
@@ -76,6 +76,7 @@
 #include "supported/Solaris.hpp"
 #include "supported/SpaceInvaders.hpp"
 #include "supported/StarGunner.hpp"
+#include "supported/Surround.hpp"
 #include "supported/Tennis.hpp"
 #include "supported/Tetris.hpp"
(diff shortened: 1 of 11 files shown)


src/games/Roms.cpp
```diff
@@ -185,3 +187,11 @@ RomSettings *buildRomRLWrapper(const std::string &rom) {
     return NULL;
 }
 
+// Taken from Xitari.
```
mcmachado: Don't you have to add a comment at the header, like I did for the lives signal?

mcmachado:
I didn't test it, but it LGTM.
I thought you were going to PR to my repository, not this one...

src/games/Roms.cpp
mgbellemare: Good idea.

nikonikolov:
Was Surround added eventually? I cannot seem to find it in the master branch.

PR #190, opened 2017-03-19
Add Kaboom! rom

nczempin:
I figured out where and how [Kaboom!](https://en.wikipedia.org/wiki/Kaboom!_(video_game)) stores the score and the number of "lives" (buckets), with help from the Stella debugger and this disassembly: http://www.bjars.com/source/Kaboom.asm.

When the score reaches 999,999 points, the game freezes, so I used that as a terminal condition (in addition to the regular one in which you have no more lives left).

### DIFF
--- a/src/games/module.mk
+++ b/src/games/module.mk
@@ -39,6 +39,7 @@ MODULE_OBJS := \
 	src/games/supported/IceHockey.o \
 	src/games/supported/JamesBond.o \
 	src/games/supported/JourneyEscape.o \
+	src/games/supported/Kaboom.o \
 	src/games/supported/Kangaroo.o \
 	src/games/supported/Koolaid.o \
 	src/games/supported/KeystoneKapers.o \
--- a/src/games/supported/Kaboom.cpp
+++ b/src/games/supported/Kaboom.cpp
@@ -1,0 +1,99 @@
+/* *****************************************************************************
+ * A.L.E (Arcade Learning Environment)
+ * Copyright (c) 2009-2013 by Yavar Naddaf, Joel Veness, Marc G. Bellemare and 
+ *   the Reinforcement Learning and Artificial Intelligence Laboratory
+ * Released under the GNU General Public License; see License.txt for details. 
+ *
+ * Based on: Stella  --  "An Atari 2600 VCS Emulator"
+ * Copyright (c) 1995-2007 by Bradford W. Mott and the Stella team
+ *
+ * *****************************************************************************
+ */
+#include "Kaboom.hpp"
+
+#include "../RomUtils.hpp"
+#include <iostream>
+
+KaboomSettings::KaboomSettings() {
+    reset();
+}
+
+
+/* create a new instance of the rom */
+RomSettings* KaboomSettings::clone() const {
+    
+    RomSettings* rval = new KaboomSettings();
+    *rval = *this;
+    return rval;
+}
+
+
+/* process the latest information from ALE */
+void KaboomSettings::step(const System& system) {
+
+    // update the reward
+    reward_t score = getDecimalScore(0xA5, 0xA4, 0xA3, &system);
+    m_reward = score - m_score;
+    m_score = score;
+
+    // update terminal status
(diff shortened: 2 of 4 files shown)


mgbellemare:
Hi @nczempin , thanks for the suggestions & updates! I'll sort these out as soon as I get the chance. Re: surround, iirc the extra select + reset actions weren't directly implementable, which has delayed resolving this issue.

src/games/module.mk
```diff
@@ -39,6 +39,7 @@ MODULE_OBJS := \
 	src/games/supported/IceHockey.o \
 	src/games/supported/JamesBond.o \
 	src/games/supported/JourneyEscape.o \
+	src/games/supported/Kaboom.o
```
mgbellemare: Seems to be missing a end of line \.

src/games/supported/Kaboom.cpp
```diff
@@ -0,0 +1,98 @@
...
+void KaboomSettings::step(const System& system) {
+
+    // update the reward
+    reward_t score = getDecimalScore(0xA5, 0xA4, 0xA3, &system);
+    m_reward = score - m_score;
+    m_score = score;
+
+    // update terminal status
+    int lives = readRam(&system, 0xA1);
+    m_terminal = lives == 0x0 || m_score > 999991;
```
mgbellemare: Why the particular choice of 999991, rather than == at 999999?

nczempin: I explained this in the description; it would be aliased

src/games/supported/Kaboom.cpp
```diff
@@ -0,0 +1,98 @@
...
+}
+
+
+/* is an action part of the minimal set? */
+bool KaboomSettings::isMinimal(const Action &a) const {
+
+    switch (a) {
+        case PLAYER_A_NOOP:
+        case PLAYER_A_RIGHT:
+        case PLAYER_A_LEFT:
```
mgbellemare: 'PLAYER_A_FIRE' is also needed to play the game.

nczempin: I thought that was taken care of by adding it to the starting actions in line 96

nczempin: The fire button is not needed to play the game, only to start it. I see it as outside of the scope of an agent that's trying to play the game (it has no effect on game state), but if you want me to add it back, of course I can do so.

Or maybe I'm missing something about playing this game?

mgbellemare: Yes, the button is needed to play the game. Have a play via the ALE or Stella and you'll see.

src/games/supported/Kaboom.cpp
mgbellemare: Playing the game, it looks like it correctly prevents overflows and sets the score to 999,999 and then freezes the game. Let's have it end when m_score == 999,999, since that's when the game actually ends.

src/games/supported/Kaboom.cpp
nczempin: Oh, okay, I see what you mean; after a wave you need to press it, and so I guess that "add to start actions" is not sufficient.

src/games/supported/Kaboom.cpp
nczempin: ah yes, of course, for _this_ game it's not actually needed, it would only be necessary for something like missile command that just overflows, and then it's a specific choice that might be up for debate whether to let it overflow or not (or perhaps some kind of overflow counter; that is essentially how this is handled by human world record judges).

src/games/supported/Kaboom.cpp
nczempin: Okay, fixed this, too.

nczempin:
Do I need to do anything else to let github or @mgbellemare see that I've made the changes? Somehow ask for a followup-review?

mgbellemare:
Did you push the new changes? The PR isn't showing further commits.

nczempin:
I did push the changes after doing git commit --amend, so the diff you see should reflect the net result

PR #194, opened 2017-03-28
Add Robinett's Adventure

nczempin:
[Adventure](https://en.wikipedia.org/wiki/Adventure_(Atari_2600)) now works.

Reward is +1 one for bringing the chalice to the yellow castle, -1 for getting eaten by a dragon. Only these two events end an episode.


nczempin:
Tested with 4b27f5397c442d25f0c418ccdacf1926 Adventure (1978) (Atari) (PAL) [!]

src/games/supported/Adventure.hpp
```diff
@@ -0,0 +1,75 @@
...
+ *   the Reinforcement Learning and Artificial Intelligence Laboratory
+ * Released under the GNU General Public License; see License.txt for details. 
+ *
+ * Based on: Stella  --  "An Atari 2600 VCS Emulator"
+ * Copyright (c) 1995-2007 by Bradford W. Mott and the Stella team
+ *
+ * *****************************************************************************
+ */
+#ifndef __ADVENTURE_HPP__
+#define ____ADVENTURE_HPP___HPP__
```
mgbellemare: I think you mean only one HPP__.

src/games/supported/Adventure.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  if (chalice_in_yellow_castle) {
+    m_reward = 1;
+  }
+
+  int player_status = readRam(&system, 0xE0);
+  bool player_eaten = player_status == 2;
+
+  if (player_eaten) {
+    m_reward = -1;
```
mgbellemare: Can you remove the -1 for being eaten? No other game has a negative reward for terminating the episode (most agents assume that termination is itself pretty bad).

src/games/supported/Adventure.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/* process the latest information from ALE */
+void AdventureSettings::step(const System& system) {
+
+  int chalice_status = readRam(&system, 0xB9);
+  bool chalice_in_yellow_castle = chalice_status == 0x12;
+
+  if (chalice_in_yellow_castle) {
+    m_reward = 1;
```
mgbellemare: Technically, there is no +1 for winning the episode! But let's leave this one in since otherwise we won't have any easy way of distinguishing successful trials from unsuccessful ones.

src/games/supported/Adventure.cpp
mgbellemare: FYI, the game does not actually end when eaten by a dragon. Pressing reset will restart the game, but with objects left where they are. This is in contrast with ALE, which will actually reset the whole emulator and therefore reset the object positions... but for sanity's sake, we'll keep it simple and define the episode to end when eaten by a dragon.

nczempin: Yes, and one side effect was that on the higher levels the bat could grab something that would cause the dragon to carry you through the world; I forget the details.

src/games/supported/Adventure.hpp
nczempin: Um, yes.

src/games/supported/Adventure.cpp
nczempin: About removing the -1: of course I can and will do it. On the other hand, you really want to make it even harder? I'd be very curious to see what your intrinsically motivated agent does with this game.

src/games/supported/Adventure.cpp
nczempin: Hm, not that it matters, but I don't quite understand you. There is no score in this game, but that doesn't mean there should be no reward.

In that respect it's more like chess or go, where you are only rewarded at the end, but not with a point (unless you play in a tournament). That was also my motivation for adding the -1; anyrhing else is undecided, and I guess inconclusive games (I don't know the details of the timeout that must exist, is that external to ALE or is there a step limit?) are then draws.

So, aside from the fact that you do say that we should leave it in, under what circumstance could we ever use a game that nevee hands out any reward at all? No, you don't need to answer, it's not that important, just wondering.

src/games/supported/Adventure.cpp
nczempin: And my only other concern is that openai have already merged my PR with the older  version both of this (with the -1) and Kaboom. 

I guess it's a discussion for another day, but ideally I'd like to help bringing the fork(s) back to the parent rather than let it diverge, especially on such a minor issue.

But I guess I need to take it up with them.

src/games/supported/Adventure.cpp
mgbellemare: My usual thinking is to bring as little human knowledge into the game as possible, so we can really claim generality. It's unfortunately very easy to infuse prior knowledge in RL via the reward function.  I like your chess analogy, but I'm thinking here there's no reason to treat Adventure differently -- for example, why not then add a negative reward for death in Pitfall, Montezuma's Revenge, etc? Of course, the agent designer is always free to do so (by looking at the termination signal), which is not necessarily a bad idea -- value-based agents typically clamp the value function at 0 at episode ends.

nczempin: I think we've handled the part about the negative reward elsewhere, and in my comment just above your last one I was just curious because it seemed you were suggesting that there could be a scenario in which there is no reward at all, not even on completing the mission.

Perhaps my mentioning the negative reward obscured what I was primarily interested in.

nczempin:
Good point. I know that at the higher levels you could get into a state
where the bat would somehow give you a flyover of the kingdom, although I'm
not sure there way ever a way to get out of the dragon.

Also, your comment on lives in general convinced me that right now we'd be
consistent with the rest by not treating "eaten" as anything special.

I still believe that there is a case for including more features, but I
won't argue it here and now.

On 9 Apr 2017 10:48 pm, "Marc G. Bellemare" <notifications@github.com>
wrote:

*@mgbellemare* commented on this pull request.
------------------------------

In src/games/supported/Adventure.cpp
<https://github.com/mgbellemare/Arcade-Learning-Environment/pull/194#discussion_r110552406>
:

> +
+void AdventureSettings::step(const System& system) {
+
+  int chalice_status = readRam(&system, 0xB9);
+  bool chalice_in_yellow_castle = chalice_status == 0x12;
+
+  if (chalice_in_yellow_castle) {
+    m_reward = 1;
+  }
+
+  int player_status = readRam(&system, 0xE0);
+  bool player_eaten = player_status == 2;
+
+  if (player_eaten) {
+    m_reward = -1;

FYI, the game does not actually end when eaten by a dragon. Pressing reset
will restart the game, but with objects left where they are. This is in
contrast with ALE, which will actually reset the whole emulator and
therefore reset the object positions... but for sanity's sake, we'll keep
it simple and define the episode to end when eaten by a dragon.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mgbellemare/Arcade-Learning-Environment/pull/194#discussion_r110552406>,
or mute the thread

PR #197, opened 2017-04-03
Add Kaboom! rom

nczempin:
PRs to dev are more appropriate


mgbellemare:
Probably, but the patch to master was fine in this case too. Thanks for adding support for Kaboom!

PR #198, opened 2017-04-04
Separate build targets for the libraries and binary

shelhamer:
In my usage of ALE I only ever need the C library for the Python interface. For convenience I separated the build targets so that it isn't necessary to compile whichever of the C++ library, C library, and ALE binary are unneeded in a particular use case.

p.s. This is my first attempt at a CMake patch so please double-check that its not in error. I'd be happy to correct it if needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,9 @@ set(ALEVERSION "0.5")
 option(USE_SDL "Use SDL" OFF)
 option(USE_RLGLUE "Use RL-Glue" OFF)
 option(BUILD_EXAMPLES "Build Example Agents" ON)
+option(BUILD_CPP_LIB "Build C++ Shared Library" ON)
+option(BUILD_CLI "Build ALE Command Line Interface" ON)
+option(BUILD_C_LIB "Build ALE C Library (needed for Python interface)" ON)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wunused -fPIC -O3 -fomit-frame-pointer -D__STDC_CONSTANT_MACROS")
 add_definitions(-DHAVE_INTTYPES)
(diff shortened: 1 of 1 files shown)


mgbellemare:
Thanks for the update! Was there a particular reason, other than cosmetics, for separating the two builds -- given that both still require building all of the source?

CMakeLists.txt
```diff
@@ -6,6 +6,9 @@ set(ALEVERSION "0.5")
 option(USE_SDL "Use SDL" OFF)
 option(USE_RLGLUE "Use RL-Glue" OFF)
 option(BUILD_EXAMPLES "Build Example Agents" ON)
+option(BUILD_CPP_LIB "Build C++ Shared Library" ON)
+option(BUILD_CLI "Build ALE Command Line Interface" ON)
```
mgbellemare: Is this flag used at all?

CMakeLists.txt
```diff
@@ -6,6 +6,9 @@ set(ALEVERSION "0.5")
 option(USE_SDL "Use SDL" OFF)
 option(USE_RLGLUE "Use RL-Glue" OFF)
 option(BUILD_EXAMPLES "Build Example Agents" ON)
+option(BUILD_CPP_LIB "Build C++ Shared Library" ON)
```
mgbellemare: Setting BUILD_CPP_LIB off but leaving BUILD_EXAMPLES on will fail. Can you add some code that errors in this case?

CMakeLists.txt
shelhamer: sorry—I accidentally switched flags part-way through the patch. Fixed in latest commit by `BUILD_ALE_BIN` -> `BUILD_CLI`

CMakeLists.txt
shelhamer: Thanks for catching that. I've added a rule to force building the C++ lib when the examples are to be built.

shelhamer:
> Was there a particular reason, other than cosmetics, for separating the two builds -- given that both still require building all of the source?

My Python-only use case skips the build of the C++ library and the command line interface/ALE binary. It's not necessarily dramatic, but it does save a little time.

PR #200, opened 2017-04-10
Issue/#123 include md5 checksums

### LINKED ISSUE #123: Include MD5 or SHA1 checksum for all supported ROM files
Dear developers,

please could you include as a text file in the repository a list of MD5 sums or SHA1 sums of all ROMs currently supported?  For example, you can just run `md5sum *.bin` in your directory of working ROMs to create such a file.

This would make it easier to collect/rename all ROM files.  For example, for some games there are multiple versions (homebrew Tetris for example has the original .a26 version and an updated one, other games have PAL and NTSC versions), and this would make it easy.

Thank you for considering and your excellent work.

Sebastian

nczempin:
Fixes #123


nczempin:
oh, maybe I should have done this on the add-md5 branch, sry.

src/ale_interface.cpp
```diff
@@ -111,7 +141,32 @@ ALEInterface::ALEInterface(bool display_screen) {
...
+}
+
+bool is_illegal(char c) {
+  return !(std::isalnum(c) || c == '_');
+}
+
+string ALEInterface::find_rom(const string& md5) {
+  // TODO these values should either come from an external text file or be part of the RomSettings subclasses
+  // TODO building a dictionary would be more elegant, but overkill for so few strings
+  // TODO this can be done much more elegantly with C++11
```
mgbellemare: I'm thinking this should go into the dev branch, which will be C++11. Would that work?

src/ale_interface.cpp
```diff
@@ -111,7 +141,32 @@ ALEInterface::ALEInterface(bool display_screen) {
...
+  std::ifstream ss("md5.txt");
```
mgbellemare: Is there a better place than here to encode the md5.txt file?

src/ale_interface.cpp
```diff
@@ -111,7 +141,32 @@ ALEInterface::ALEInterface(bool display_screen) {
...
 
-ALEInterface::~ALEInterface() {}
+ALEInterface::~ALEInterface() {
+}
+
+bool is_illegal(char c) {
+  return !(std::isalnum(c) || c == '_');
+}
+
+string ALEInterface::find_rom(const string& md5) {
```
mgbellemare: std::string

src/ale_interface.cpp
```diff
@@ -111,7 +141,32 @@ ALEInterface::ALEInterface(bool display_screen) {
...
+  // TODO these values should either come from an external text file or be part of the RomSettings subclasses
+  // TODO building a dictionary would be more elegant, but overkill for so few strings
+  // TODO this can be done much more elegantly with C++11
+  std::ifstream ss("md5.txt");
+  std::string item;
+  string rom_candidate = "";
```
mgbellemare: std::string

src/ale_interface.cpp
```diff
@@ -66,32 +82,46 @@ void ALEInterface::createOSystem(std::auto_ptr<OSystem> &theOSystem,
...
+    const Properties properties = theOSystem->console().properties();
+    const std::string name = properties.get(Cartridge_Name);
+    const std::string md5 = properties.get(Cartridge_MD5);
+    Logger::Info << "Cartridge_name: " << name << std::endl;
+    Logger::Info << "Cartridge_MD5: " << md5 << std::endl;
+    string rom_candidate = find_rom(md5);
+    if (rom_candidate == "") {
+      Logger::Warning << "Warning. Possibly unsupported ROM." << std::endl;
+    }
     Logger::Info << "Running ROM file..." << std::endl;
```
mgbellemare: Test also that rom_candidate == romfile?

src/ale_interface.cpp
```diff
@@ -66,32 +82,46 @@ void ALEInterface::createOSystem(std::auto_ptr<OSystem> &theOSystem,
...
+    Logger::Error << "ROM file " << romfile << " not found." << std::endl;
     exit(1);
-  } else if (theOSystem->createConsole(romfile))  {
+  } else if (theOSystem->createConsole(romfile)) {
+    const Properties properties = theOSystem->console().properties();
+    const std::string name = properties.get(Cartridge_Name);
+    const std::string md5 = properties.get(Cartridge_MD5);
+    Logger::Info << "Cartridge_name: " << name << std::endl;
+    Logger::Info << "Cartridge_MD5: " << md5 << std::endl;
+    string rom_candidate = find_rom(md5);
```
mgbellemare: std::string

src/ale_interface.cpp
```diff
@@ -66,32 +82,46 @@ void ALEInterface::createOSystem(std::auto_ptr<OSystem> &theOSystem,
...
+    if (rom_candidate == "") {
```
mgbellemare: perhaps better: ```if (string.empty())```

src/ale_interface.cpp
```diff
@@ -123,15 +178,44 @@ void ALEInterface::loadROM(string rom_file = "") {
...
+        << "'. Trying alternative method..." << std::endl;
+
+    std::string s = "Unknown";
+    std::string s1 = s.substr(0, s.find("("));
+    s1.erase(std::remove_if(s1.begin(), s1.end(), is_illegal), s1.end());
+    std::cout << s1 << std::endl;
+
+    const Properties properties = theOSystem->console().properties();
+    const std::string md5 = properties.get(Cartridge_MD5); //TODO there shouldn't be a need to do this twice
+    string rom_candidate = find_rom(md5);
```
mgbellemare: const std::string

src/ale_interface.cpp
```diff
@@ -123,15 +178,44 @@ void ALEInterface::loadROM(string rom_file = "") {
...
+    if (rom_candidate != "") {
```
mgbellemare: Preferable is to raise exception if rom_candidate doesn't match requested rom. See also above re: rom_candidate.empty()

src/ale_interface.cpp
nczempin: definitely better

src/ale_interface.cpp
nczempin: I'll give it a go

nczempin:
This currently doesn't have a high priority for me; it was more or less rushed to give someone who may want it the opportunity to use some md5 checking, but it needs a bit of massaging.

src/ale_interface.cpp
nczempin: the comment is partly outdated; originally I just hardcoded the strings in the .cpp file.

src/ale_interface.cpp
nczempin: Hm. This code is not really necessary for the md5 stuff; it just happened that I was working in the vicinity and I got segfaults when using roms from the wild that didn't match the naming convention.

It's probably best to reduce this PR to just the MD5 part, then we can discuss how to handle "unexpected ROM" that goes beyond using the correct name with an "unsupported" ROM there.

Raising exception would just be a slightly friendlier version than the segfault, and it would force users to rename all there ROMs for no reason that I can think of: The ROMs can all be identified via their MD5 checksums, it should be completely irrelevant what the filenames are. AFAIR the previous method even had issues when using ROMs outside the current directory, although that may have been local to my own modified version.

nczempin: or did you mean throw an exception instead of exit(1)? that part is not visible in the review comment

nczempin: or did you mean throw an exception instead of the exit(1)?

nczempin:
I've isolated the code that is only concerned with the md5 warning to PR #202

PR #205, opened 2017-05-31
fix src_size bug

skylian:
`screen_size` will be multiplied by 3 inside the function [applyPaletteRGB](https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/common/ColourPalette.cpp#L89).


ppwwyyxx:
I just found the same problem and mentioned in #161

mgbellemare:
Thank for the fix. Can you separate the max pooling code from fix? Max pooling as given here is neat, but not exactly what used in e.g. DQN -- because averaged colours get mapped back to NTSC. So it might be worthwhile to discuss a little more how to get about this.

skylian:
Sorry. I didn't know I was at master branch.

PR #209, opened 2017-08-23
Minimal definition of interfaces to support modes and difficulties.

mcmachado:
@mgbellemare, I created a first minimal pull request for this feature. In this first PR all I'm doing is defining types and the interface I want to make accessible for the end user. If you are ok with it, I'll proceed to implement the functionalities, to update the manual, and to implement the Python interface as well.

### DIFF
--- a/doc/examples/Makefile.sharedLibraryWithModes
+++ b/doc/examples/Makefile.sharedLibraryWithModes
@@ -1,0 +1,31 @@
+USE_SDL := 0
+
+# This will likely need to be changed to suit your installation.
+ALE := ../..
+
+FLAGS := -I$(ALE)/src -I$(ALE)/src/controllers -I$(ALE)/src/os_dependent -I$(ALE)/src/environment -I$(ALE)/src/external -L$(ALE)
+CXX := g++
+FILE := sharedLibraryInterfaceWithModesExample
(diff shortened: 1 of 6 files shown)


doc/examples/Makefile.sharedLibraryWithModes
```diff
@@ -0,0 +1,31 @@
+USE_SDL := 0
```
mgbellemare: Are people using these Makefiles?

src/ale_interface.cpp
```diff
@@ -266,6 +266,36 @@ reward_t ALEInterface::act(Action action) {
   return reward;
 }
 
+// Returns the vector of modes available for the current game.
+// This should be called only after the rom is loaded.
+ModeVect ALEInterface::getAvailableModes(){
```
mgbellemare: Space between () and { please (also applies to other methods below).

doc/examples/Makefile.sharedLibraryWithModes
mcmachado: I don't know. But I don't see why it hurts to keep them.

mcmachado:
Closed because I'll work on dev.

PR #212, opened 2017-08-26
Accessing other events inside Stella and allowing ALE to change RAM values

mcmachado:
The last set of functionalities I need before actually implementing the mode and difficulty changes are the abilities to:
 - deal with other events supported in Stella but not previously supported in the ALE (Setting difficulty level and pressing select).
 - write to RAM so I can set the game mode.

I believe my next PR, after these two features, will finalize the implementation of the desired functionalities (stella_environment and RomSettings). Then it will be a matter of updating the code in all games.

### DIFF
--- a/src/environment/ale_state.cpp
+++ b/src/environment/ale_state.cpp
@@ -475,7 +475,7 @@ void ALEState::setActionJoysticks(Event* event, int player_a_action, int player_
  * ***************************************************************************/
 void ALEState::resetKeys(Event* event) {
     event->set(Event::ConsoleReset, 0);
-    event->set(Event::ConsoleSelect,0);
+    event->set(Event::ConsoleSelect, 0);
     event->set(Event::JoystickZeroFire, 0);
     event->set(Event::JoystickZeroUp, 0);
     event->set(Event::JoystickZeroDown, 0);
--- a/src/environment/ale_state.cpp
+++ b/src/environment/ale_state.cpp
@@ -259,6 +259,19 @@ void ALEState::applyActionPaddles(Event* event, int player_a_action, int player_
   }
 }
 
+void ALEState::pressSelect(Event* event) {
+  resetKeys(event);
+  event->set(Event::ConsoleSelect, 1);
+}
+
+void ALEState::setDifficulty(Event* event, unsigned mask) {
+  resetKeys(event);
+  event->set(Event::ConsoleLeftDifficultyA, mask & 1);
+  event->set(Event::ConsoleLeftDifficultyB, !(mask & 1));
+  event->set(Event::ConsoleRightDifficultyA, (mask & 2) >> 1);
+  event->set(Event::ConsoleRightDifficultyB, !((mask & 2) >> 1));
+}
+
 void ALEState::setActionJoysticks(Event* event, int player_a_action, int player_b_action) {
   // Reset keys
   resetKeys(event);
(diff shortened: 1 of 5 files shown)


src/games/RomUtils.hpp
```diff
@@ -22,6 +22,10 @@ class System;
 // reads a byte at a memory location between 0 and 1023
 extern int readRam(const System* system, int offset);
 
+// writes a byte at a memory location between 0 and 128,
+// this method is required to allow one to set a game mode
+extern void writeRam(System* system, int offset, unsigned char value);
```
mgbellemare: I thought modes were set by repeatedly pressing select? Are you thinking it's easier to directly write to RAM?

src/environment/ale_state.cpp
```diff
@@ -259,6 +259,19 @@ void ALEState::applyActionPaddles(Event* event, int player_a_action, int player_
   }
 }
 
+void ALEState::pressSelect(Event* event) {
+  resetKeys(event);
+  event->set(Event::ConsoleSelect, 1);
+}
+
+void ALEState::setDifficulty(Event* event, unsigned mask) {
```
mgbellemare: unsigned should be unsigned int. Is there a reason to use a mask, rather than two booleans or integers (one per player)? With a mask it won't be obvious what the function arguments do.

src/environment/ale_state.cpp
```diff
@@ -259,6 +259,19 @@ void ALEState::applyActionPaddles(Event* event, int player_a_action, int player_
...
+  resetKeys(event);
+  event->set(Event::ConsoleLeftDifficultyA, mask & 1);
```
mgbellemare: If I understand correctly, this scheme assumes that 'event' remains the same throughout -- i.e. the difficulty bits never get cleared except through this method. Is this going to be an issue for serialization?

src/games/RomUtils.cpp
```diff
@@ -30,6 +30,12 @@ int readRam(const System* system, int offset) {
 }
 
 
+/* writes a byte at a memory location between 0 and 128 */
+void writeRam(System* system, int offset, unsigned char value) {
+
```
mgbellemare: No whitespace here (or be consistent across functions).

src/games/RomUtils.hpp
mcmachado: I was following the previous pull requests and this was in there. If I remember correctly, we used this initially for debugging and it was left in there. I'll remove it for now. I don't think I'll need it.

src/games/RomUtils.cpp
mcmachado: I'm doing my best, but since it is different from what he had before, sometimes things something slips in. Thanks for pointing that out.

src/environment/ale_state.cpp
mcmachado: The mask makes some of the code more transparent. This is because the difficulty switch was a 0 and 1, and when we request getDifficulties it returns that.

src/environment/ale_state.cpp
mcmachado: To be honest, I can't say for sure.
I think it would be better if we went through with this now and then we check serialization, for example. Since more has to be changed in order to get this feature, I'd rather do this and then change the code accordingly if serialization becomes a problem.

mgbellemare:
Please add issue, otherwise good to merge.

PR #214, opened 2017-08-29
Main bulk of mode and difficulty switch

mcmachado:
@mgbellemare, this is the actual implementation of the mode and difficulty switch. I implemented these features in just one game for now (Freeway) to show how they should look like. As soon as we merge this PR I'll start working on the games.

**Important comment:**
There is a circular dependency between RomSettings and StellaEnvironment. I know this is far from ideal. However, I couldn't really find a way to solve it. We want to define the methods in RomSettings to force all Roms to implement them. However, RomSettings do not have the required data so we can implement everything on it.

My dilemma is the following: if I put all the information inside StellaEnvironment, it is not going to be in the super class of each game. And I'd always have to pass the environment as a parameter to each function (which in fact wouldn't solve the circular dependency). I can't put everything inside RomSettings because I need to call several functions implemented in StellaEnvironment.

My opinion about this is the following: if you have a suggestion on how to do it in a more principled way, I can try to spend 1 or 2 hours doing so. However, in case a major refactoring is required, I'd suggest we merge this, finalize the PR, and create an issue to make the code nicer in the future. What do you say?


src/ale_interface.cpp
```diff
@@ -269,31 +269,32 @@ reward_t ALEInterface::act(Action action) {
...
+  return romSettings->getAvailableDifficulties();
 }
 
 // Sets the difficulty of the game.
 // The difficulty must be an available mode.
 // This should be called only after the rom is loaded.
 void ALEInterface::setDifficulty(difficulty_t m) {
-  throw std::runtime_error("This feature is not available yet.");
+  DifficultyVect available = romSettings->getAvailableDifficulties();
+  if(find(available.begin(), available.end(), m) != available.end()){
```
mgbellemare: Space between ) and {.

mgbellemare: Is there a reason to not push the check inside environment->setDifficulty?

src/environment/ale_state.cpp
```diff
@@ -12,11 +12,16 @@
...
+#include "../emucore/Deserializer.hxx"
+#include "../emucore/Serializer.hxx"
 #include "../common/Constants.h"
-using namespace std;
+#include "../games/RomSettings.hpp"
 
+#include <sstream>
 #include <stdexcept>
 
+using namespace std;
```
mgbellemare: We should get rid of the using namespace std.

src/environment/ale_state.hpp
```diff
@@ -21,9 +21,10 @@
 #include "../emucore/OSystem.hxx"
 #include "../emucore/Event.hxx"
 #include <string>
-#include "../games/RomSettings.hpp"
 #include "../common/Log.hpp"
 
+class RomSettings;
```
mgbellemare: Where is this needed? It's bad to use a class forward declaration, but worse to put it in the header file.

src/ale_interface.cpp
```diff
@@ -269,31 +269,32 @@ reward_t ALEInterface::act(Action action) {
...
-  return placeholder;
+  return romSettings->getAvailableDifficulties();
 }
 
 // Sets the difficulty of the game.
 // The difficulty must be an available mode.
 // This should be called only after the rom is loaded.
 void ALEInterface::setDifficulty(difficulty_t m) {
-  throw std::runtime_error("This feature is not available yet.");
+  DifficultyVect available = romSettings->getAvailableDifficulties();
```
mgbellemare: Is "difficulty" really a single value? Is this how we want to think of it?

src/environment/stella_environment.cpp
```diff
@@ -195,6 +203,26 @@ bool StellaEnvironment::isTerminal() const {
...
+  m_state.pressSelect(event);
+  for (size_t t = 0; t < num_steps; t++) {
+    m_osystem->console().mediaSource().update();
+  }
+  processScreen();
+  processRAM();
+  emulate(PLAYER_A_NOOP, PLAYER_B_NOOP);
+}
+
+void StellaEnvironment::setDifficulty(difficulty_t mask) {
```
mgbellemare: This 'mask' variable name doesn't do it for me. Can we call it 'difficulty' or 'value'? 'Mask' suggests that we need to know the internal details of the integer, in which case I would split it into two arguments (left_difficulty and right_difficulty).

src/environment/stella_environment.cpp
```diff
@@ -195,6 +203,26 @@ bool StellaEnvironment::isTerminal() const {
...
+  Event* event = m_osystem->event();
+  m_state.setDifficulty(event, mask);
+  m_osystem->console().mediaSource().update();
+  processScreen();
+  processRAM();
+  m_difficulty = mask;
```
mgbellemare: Is there a reason for story m_difficulty in StellaEnvironment, rather than ALEState?

src/environment/stella_environment.hpp
```diff
@@ -30,6 +30,8 @@
 
 #include <stack>
 
+class StellaEnvironment;
```
mgbellemare: This forward declaration is spurious, no?

src/environment/stella_environment.hpp
```diff
@@ -64,6 +66,18 @@ class StellaEnvironment {
...
+    void softReset();
+
+    /** Keep pressing the console select button for a given amount of time*/
+    void pressSelect(size_t num_steps = 1);
+
+    /** Set the difficulty according to the mask.
+      * If the first bit is 1, then it will put the left difficulty switch to A (otherwise leave it on B)
+      * If the second bit is 1, then it will put the right difficulty switch to A (otherwise leave it on B)
+      */
+    void setDifficulty(difficulty_t mask);
```
mgbellemare: See comments above.

src/games/RomSettings.hpp
```diff
@@ -33,19 +33,34 @@
...
 
     // reset
-    virtual void reset() = 0;
+    virtual void reset(System& system, StellaEnvironment& environment) {
+        //ignore parameters by default
+        reset();
+    }
+
+    // reset
+    virtual void reset(){};
```
mgbellemare: This used to be abstract. Is there a reason not to keep it so?

src/games/RomSettings.hpp
```diff
@@ -33,19 +33,34 @@
...
+class RomSettings {
+
+public:
+    RomSettings();
+
     virtual ~RomSettings() {}
 
     // reset
-    virtual void reset() = 0;
+    virtual void reset(System& system, StellaEnvironment& environment) {
```
mgbellemare: Please put code in the cpp file.

src/games/RomSettings.hpp
```diff
@@ -86,6 +101,25 @@ struct RomSettings {
     // Returns a list of actions that are required to start the game.
     // By default this is an empty list.
     virtual ActionVect getStartingActions();
+
+    // Returns a list of mode that the game can be played in. 
+    // By default, there is only one available mode.
+    virtual ModeVect getAvailableModes() {
+        return ModeVect(1, 0);
```
mgbellemare: Code in .cpp.

src/games/RomSettings.hpp
```diff
@@ -86,6 +101,25 @@ struct RomSettings {
...
+    };
+
+    // Set the mode of the game. The given mode must be
+    // one returned by the previous function.
+    virtual void setMode(game_mode_t, System &system, StellaEnvironment& environment);
+
+    // Returns a list of difficulties that the game can be played in.
+    // By default, there is only one available difficulty.
+    virtual DifficultyVect getAvailableDifficulties() {
+        return DifficultyVect(1, 0);
```
mgbellemare: Code in cpp.

src/games/RomSettings.hpp
```diff
@@ -33,19 +33,34 @@
...
+#include <stdexcept>
+
 #include "../common/Constants.h"
 #include "../emucore/Serializer.hxx"
 #include "../emucore/Deserializer.hxx"
+#include "../environment/stella_environment.hpp"
 
 class System;
-
+class RomSettings;
```
mgbellemare: At least one of these is unnecessary.

src/games/RomSettings.cpp
```diff
@@ -17,6 +17,8 @@
  */
 #include "RomSettings.hpp"
 
+RomSettings::RomSettings() : m_mode(0) {}
```
mgbellemare: The m_mode variable doesn't seem to be used. Will it be used?

Unless it's needed here, we should push it to be individual ROMSettings subclasses for consistent with m_score, m_reward, etc.

src/games/Roms.hpp
```diff
@@ -14,8 +14,7 @@
 
 #include <string>
 
-struct RomSettings;
-
+class RomSettings;
```
mgbellemare: Why is this needed?

src/games/supported/Freeway.cpp
```diff
@@ -16,7 +16,9 @@
 
 FreewaySettings::FreewaySettings() {
 
-    reset();
+    m_reward   = 0;
```
mgbellemare: IF we keep the reset(System, Environment) function (see below):

We shouldn't duplicate code, especially because a few roms with have non-trivial values (e.g. Pitfall! sets score to 2000, and lives data also). A few possibilities:

- Keep the existing reset() method, and call the new one reset_and_set_mode()
- add a reset_variables() method that wraps the setting of values
- OR, See below for a preferred solution

src/games/supported/Freeway.cpp
```diff
@@ -74,11 +76,12 @@ bool FreewaySettings::isMinimal(const Action &a) const {
 
 
 /* reset the state of the game */
-void FreewaySettings::reset() {
+void FreewaySettings::reset(System& system, StellaEnvironment& environment) {
     
     m_reward   = 0;
     m_score    = 0;
     m_terminal = false;
+    setMode(m_mode, system, environment);
```
mgbellemare: m_mode should really be stored in StellaEnvironment with m_difficulty (or even better, ALEState). RomSettings was designed to read off the current state of the game from the emulator core. I'll propose a more complete solution in the discussion.

src/games/supported/Freeway.cpp
```diff
@@ -95,3 +98,38 @@ void FreewaySettings::loadState(Deserializer & ser) {
   m_terminal = ser.getBool();
 }
 
+// returns a list of mode that the game can be played in
+ModeVect FreewaySettings::getAvailableModes() {
+    ModeVect modes(8);
+    for(unsigned i = 0; i < 8; i++) {
```
mgbellemare: unsigned int. If only this was Python!

```return [ i for i in range(8) ]```

Also space between for and (.

Also, tie for the loop to sizeof(modes) or equivalent?

File: src/games/supported/Freeway.cpp [hunk 19]
```diff
@@ -95,3 +98,38 @@ void FreewaySettings::loadState(Deserializer & ser) {
...
+        modes[i] = i;
+    }
+    return modes;
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void FreewaySettings::setMode(game_mode_t m, System &system, StellaEnvironment& environment) {
+    if(m < 8) { /*m >= 0 is implicit, since m is an unsigned int*/
```
mgbellemare: 8 shouldn't be hardcoded also. Maybe a static const int at the beginning of the file?

src/games/supported/Freeway.cpp
```diff
@@ -95,3 +98,38 @@ void FreewaySettings::loadState(Deserializer & ser) {
...
+        }
+        //reset the environment to apply changes.
+        environment.softReset();
+    } else {
+        throw std::runtime_error("This mode doesn't currently exist for this game");
+    }
+ }
+ 
+DifficultyVect FreewaySettings::getAvailableDifficulties(){
+    DifficultyVect diff;
```
mgbellemare: I think in C++11 you can do diff ({0, 1}).

src/environment/stella_environment.cpp
```diff
@@ -195,6 +203,26 @@ bool StellaEnvironment::isTerminal() const {
      m_state.getEpisodeFrameNumber() >= m_max_num_frames_per_episode));
 }
 
+void StellaEnvironment::pressSelect(size_t num_steps) {
+  Event* event = m_osystem->event();
```
mgbellemare: Pass the Event* directly to m_state.pressSelect()

src/environment/stella_environment.cpp
```diff
@@ -195,6 +203,26 @@ bool StellaEnvironment::isTerminal() const {
...
+  m_state.pressSelect(event);
+  for (size_t t = 0; t < num_steps; t++) {
+    m_osystem->console().mediaSource().update();
+  }
+  processScreen();
```
mgbellemare: Are the process() calls needed?

src/environment/stella_environment.cpp
```diff
@@ -195,6 +203,26 @@ bool StellaEnvironment::isTerminal() const {
...
+  processRAM();
+  emulate(PLAYER_A_NOOP, PLAYER_B_NOOP);
```
mgbellemare: Comparing to softReset(), are we missing a call to `incrementFrame()` here?

mgbellemare:
Thanks Marlos! Here are my two main thoughts after reading this through:

- Storing the mode, and setting it at the beginning of an episode, should be done in StellaEnvironment. RomSettings::reset() isn't meant to have side-effects.

- IIUC, the only dependency on StellaEnvironment are at the point where we press select() and reset() from the mode-setting code. This is custom per-game, so we should probably resign ourselves to needing these two function calls there. Here's my suggestion:
  * Write two functors, soft_reset_functor() and select_functor();
  * When called (I.e., their operator() method is invoked) these functors perform the required operation;
  * To do so, they need to store or wrap some functionality from StellaEnvironment;
  * Pass these to RomSettings.

I.e. we effectively want to implement some form of closure. It looks like C++11 has some of this functionality -- do you want to take a look at lambda closures in C++11? If this isn't available, we can write our own fairly easily. This will get rid of the explicit circular dependency, which we should avoid at all costs.

src/ale_interface.cpp
mcmachado: Fixed.

src/environment/stella_environment.hpp
mcmachado: You are right. Fixed.

src/environment/ale_state.cpp
mcmachado: Fixed.

src/games/RomSettings.hpp
mcmachado: Fixed.

src/games/RomSettings.hpp
mcmachado: Fixed.

src/games/RomSettings.hpp
mcmachado: Fixed.

src/games/Roms.hpp
mcmachado: Fixed.

src/games/RomSettings.hpp
mcmachado: Fixed.

src/environment/stella_environment.cpp
mcmachado: Well, I was using it inside `StellaEnvironment::reset()`. I moved it to `ALEState` and I dropped the line `setDifficulty(m_difficulty);`. I believe I don't have to set `m_difficulty` inside `ALEState` anymore since it is now in `ALEState`. When I reset the `StellaEnvironment` it will be preserved since it will be already in `ALEState`.

@mgbellemare do you agree?

src/environment/stella_environment.hpp
mcmachado: I moved `m_difficulty` to `ale_state`, but I still do a couple of things inside the `StellaEnvironment` after I set the difficulty, namely:

`m_osystem->console().mediaSource().update();`
`processScreen();`
`processRAM();`

Do you want me to get rid of this whole function?

src/games/supported/Freeway.cpp
mcmachado: Fixed.

src/games/supported/Freeway.cpp
mcmachado: Fixed.

src/ale_interface.cpp
mcmachado: I'm not sure I understand your question here. There is a right and a left difficulty switch in the console. Therefore, 4 possible difficulties (2 bits). What do you mean by "want to think of it"?

src/games/RomSettings.cpp
mcmachado: Fixed.

src/environment/ale_state.hpp
mcmachado: Well, basically, I can't manage to get rid of it. It is being used, in `stella_environment.hpp`, on the functions `load`,  `save`, and in the constructor. I couldn't find a way to get rid of this easily. Am I missing something? There's a long time since I did anything in C++.

src/environment/stella_environment.cpp
mcmachado: I believe I did what you want.

src/environment/stella_environment.cpp
mcmachado: Honestly, I'm not sure. Do you have any idea how to find that out? The question is in fact whether a frame is necessarily updated once we change the difficulty. I googled it but I couldn't find anything. I am just following @alcinos implementation. What do you want me to do?

src/environment/stella_environment.cpp
mcmachado: After the `process()` is called in `pressSelect` we call `emulate`, which call the `process()` again. So, maybe not in `pressSelect`? However, we don't call `emulate` in `setDifficulty`. Honestly, I don't know. Any suggestions on how I can figure this out for sure?

src/games/supported/Freeway.cpp
mcmachado: It seems so, however, I tried to do:

`DifficultyVect diff = {0, 1};`

and I got the following error:
` error: non-aggregate type 'DifficultyVect' (aka 'vector<unsigned int>') cannot be initialized with an initializer list`

It could be because the `CMakeLists` file is not actually using c++11? When you merged the code, is there any chance not everyone will compile with c++11, otherwise I don't know what it is.

mcmachado:
Thanks for your comments, Marc!

Tomorrow I'll get back to this PR. The main thing now is the circular dependency. I'll see if I can do what you suggested, and how long it will take. If possible, take a look at the changes I pushed, just to ensure that the two things missing are:
1. Circular dependency,
2. Difficulty mask.

src/ale_interface.cpp
mgbellemare: It seems more natural to me to think of the difficulty settings as "Player 1 Difficulty" and "Player 2 Difficulty", rather than 4 separate choices. Let me ask a different question first: when is "available difficulties" anything else than the (A, A), (A, B), (B, A), (B, B) ?

src/environment/ale_state.hpp
mgbellemare: This used to #include RomSettings.hpp, which would avoid the forward declaration. So you must have done this to break the circular dependency RomSettings -> StellaEnvironment -> RomSettings. Hopefully if we break that we can revert to `#include "../games/RomSettings.hpp"`

src/environment/stella_environment.cpp
mgbellemare: The frame number inside ALEState is used to determine maximum play length. Should the select button count towards that? Probably, because some games (although we don't support them currently) use select as an action.

src/environment/stella_environment.hpp
mgbellemare: So setDifficulty() really is "toggle the difficulty switch and take a step in the environment", right? No, that seems like the right place for the remaining code.

src/games/supported/Freeway.cpp
mgbellemare: Very possibly missing `-std=c++11` in the g++ flags. I'm surprised the code itself compiling, given that I switched on a few other C++11 specific pieces of code.

src/ale_interface.cpp
mcmachado: Marc, there are 4 difficulties. They are not tied to a specific player. For example, Venture, has 4 difficulties ([manual](https://atariage.com/manual_html_page.php?SoftwareLabelID=576)). Of course, if you want, I can change the mask to be something else, but I just want to understand what you have in mind first.

```
   Skill          Switch
   Level          Setting
     1         left B/right B
     2         left B/right A
     3         left A/right B
     4         left A/right A
```

src/environment/stella_environment.cpp
mcmachado: I changed `mask` to `value`.

src/games/supported/Freeway.cpp
mcmachado: I added the `-std=c+11` flag to `CMakeLists.txt` and now I'm able to initialize the vector the way you want.

src/environment/stella_environment.cpp
mcmachado: I added `m_state.incrementFrame();`to the end of this function.

src/ale_interface.cpp
mgbellemare: I see, that makes a lot more sense then. Do you want to put this table somewhere as a comment?

src/games/supported/Freeway.hpp
```diff
@@ -66,11 +66,24 @@ class FreewaySettings : public RomSettings {
...
+        // returns a list of difficulties that the game can be played in
+        // in this game, there are 2 available difficulties
+        DifficultyVect getAvailableDifficulties();
+
+    private:
         bool m_terminal;
         reward_t m_reward;
         reward_t m_score;
+        game_mode_t m_mode;
+        static const unsigned int m_num_modes = 8;
```
mgbellemare: Can you use a const method instead, numModes(), to match rom()?

src/games/supported/Freeway.cpp
```diff
@@ -95,3 +99,36 @@ void FreewaySettings::loadState(Deserializer & ser) {
...
+            mode = readRam(&system, 0x80);
+        }
+        //reset the environment to apply changes.
+        environment.softReset();
+    } else {
+        throw std::runtime_error("This mode doesn't currently exist for this game");
+    }
+ }
+ 
+DifficultyVect FreewaySettings::getAvailableDifficulties(){
```
mgbellemare: Space between () and {.

mgbellemare:
I'm going to push the fix to the circular dependency now. Replying to your TODO list, two more things:

1. Can you also fix RomSettings::reset() so that it doesn't call setMode()? RomSettings::reset() should only clear the values inside the class, not act on the environment. We're making an exception for setMode() because it's too messy to do it otherwise.

2. The new example, sharedLibraryExampleWithModes, should compile to an executable called ale-sharedLibraryExampleWithModes to match the other examples.

src/ale_interface.cpp
mcmachado: Ok. I'll do that. But I'll write it as an illustrative example because, as I said, this table is from `Venture`.

mgbellemare:
I've added a branch mcmachado-up-to-date-ALE. Best is probably for you to merge my changes into your branch. It's a bit more hoop-jumping than using a forward declaration, but keeps the code sane.

src/games/supported/Freeway.cpp
mcmachado: Fixed.

src/games/supported/Freeway.hpp
mcmachado: Done.

mcmachado:
Thanks, Marc. I changed everything you asked but one thing:

> Can you also fix `RomSettings::reset()` so that it doesn't call `setMode()`? `RomSettings::reset()` should only clear the values inside the class, not act on the environment. We're making an exception for `setMode()` because it's too messy to do it otherwise.

I did this here because when we call reset (for example, when an episode is over during learning), we want to make sure we are not going to reset to the original mode. Otherwise, everyone will have to set the mode at the end of each episode. What do you think?

Aside from that, I guess all I have to do now is to look at the changes you are going to make, right?

mgbellemare:
setMode() should be inside StellaEnvironment::reset(), rather than RomSettings::reset() -- people shouldn't ever call RomSettings::reset() directly (actually, the method shouldn't be public!).

The changes are all in the branch I mentioned already.

mcmachado:
Thanks, Marc. I merged the code. All is left now is to remove `setMode` from `RomSettings`. I'll do this tomorrow and then hopefully we can merge this code.

mcmachado:
Marc: I believe I went through everything you suggested. I just finished moving `setMode` from the `reset` in `RomSettings` to the `reset` in `StellaEnvironment`. This allowed me to restore the `reset` method to what it was before, with no arguments being required.

Do you want to take a final look before merging this PR? Once we merge this PR I'll go to the next phase. I'll start pushing the changes for individual games.

src/ale_interface.hpp
```diff
@@ -97,7 +97,16 @@ class ALEInterface {
   void setMode(game_mode_t m);
 
   //Returns the vector of difficulties available for the current game.
-  //This should be called only after the rom is loaded.
+  //This should be called only after the rom is loaded. Notice
```
mgbellemare: Perfect, thanks!

mgbellemare:
One last thing: m_mode should be in the same class as m_difficulty. Can you move it out of RomSettings and put it in ALEState?

mcmachado:
Marc: I believe I've changed everything you asked for. Please let me know if you think something else has to be looked at.

src/games/supported/Freeway.cpp
```diff
@@ -110,20 +109,16 @@ ModeVect FreewaySettings::getAvailableModes() {
...
-        }
-        //reset the environment to apply changes.
-        environment->softReset();
-    } else {
-        throw std::runtime_error("This mode doesn't currently exist for this game");
+
+    // read the mode we are currently in
+    unsigned char mode = readRam(&system, 0x80);
+    // press select until the correct mode is reached
+    while (mode != m) {
```
mgbellemare: Was there a reason for removing the check `if (m < getNumModes())`? Currently this code will loop forever if an invalid m is given.

mcmachado: Yes. I changed `ale_interface.cpp`. Before I'd only call `setMode` and that was it. Now it reads like this:

```
void ALEInterface::setMode(game_mode_t m) {
  //We first need to make sure m is an available mode
  ModeVect available = romSettings->getAvailableModes();
  if(find(available.begin(), available.end(), m) != available.end()) {
    environment->setMode(m);
  } else {
    throw std::runtime_error("This mode doesn't currently exist for this game");
  }
}
```

This guarantees that `environment->setMode(m)` will only be called if the mode passed as parameter is valid. Thus, I think that check I removed was redundant.

mcmachado:
I'm merging this code. If you think the check `m < getNumModes()` is necessary, let me know. An issue would probably be enough and I'll go back to it.

PR #216, opened 2017-09-09
add std:: to fix compilation

ppwwyyxx:


mcmachado:
Can you give us some details about your problem?
What is your system? OS?

This is a fairly old file and we never had to change it to properly compile the ALE.

ppwwyyxx:
We were building https://github.com/openai/atari-py on Facebook infra based on buck.
Don't have the error log, but fixing this indeed fix the problem for us.
Mirror PR at: https://github.com/openai/atari-py/pull/19

PR #220, opened 2017-09-18
Reverting name of examples about modes

renattou:
The new name was not being ignored but git, because of its name. Also the clean rule wasn't updated correctly. Since there's no problem with the previous name, I'm just changing it back.


mgbellemare:
Thanks for the PR.

Issue #224, opened 2017-10-29
Software Optimization for ALE


pedronahum:
Hi,

I have been benchmarking the performance in Windows (https://github.com/pedronahum/Arcade-Learning-Environment), and noticed significant improvements just by flattening some multi-dimensional arrays (eg, m_rgb_ntsc, m_avg_palette, etc). 

Would you be interested in a pull request to optimize the performance of ALE for the GCC and ICC compiler? 

Regards,

mgbellemare:
Hi,

Good to hear from you, that sounds very interesting. Your fork suggests 30-35% improvements -- quite impressive. Is that mostly from flattening the arrays?

This is something we'd certainly be interested in having a look into. Could you push individual optimizations so I can have a look at them carefully?

pedronahum:
Hi,

To be honest, I didnt much do much :o) 

Flattening the arrays gave me the biggest performance bump (25-30%). Then the second contributor was data alignment. Data alignment, however, may be more complicated to generalize across compilers. 

Happy to send you individual optimizations, but maybe it is an overkill as just flattening m_rgb_ntsc, m_avg_palette and myPriorityEncoder will give you a very good performance bump. 

Happy to help, just let me know.

mgbellemare:
Let's start with the array flattening then. Thanks!

pedronahum:
Very good. In the next few days I will create an array flattening branch and send to you for your review. Thanks

PR #226, opened 2017-11-17
Array flattening

pedronahum:
Hi,

Just wanted to make sure all compiler warnings created by the flattening were addressed. As discussed in https://github.com/mgbellemare/Arcade-Learning-Environment/issues/224, flattening these arrays may increase the performance of ALE. 

Let me know if you would like me to check or test something. 

Regards,

### DIFF
--- a/src/emucore/TIA.cxx
+++ b/src/emucore/TIA.cxx
@@ -1784,7 +1784,7 @@ inline void TIA::updateFrameScanline(uInt32 clocksToUpdate, uInt32 hpos)
             enabled |= myM0Bit;
 
           myCollision |= ourCollisionTable[enabled];
-          *myFramePointer = myColor[myPriorityEncoder[(enabled | myPlayfieldPriorityAndScore) * 256 + hpos < 80 ? 0 : 1]];
+          *myFramePointer = myColor[myPriorityEncoder[ (hpos < 80 ? 0 : 1)*256 + (enabled | myPlayfieldPriorityAndScore)]];
         }
         break;  
       }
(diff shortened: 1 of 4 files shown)


mgbellemare:
Hi,

I've tested out your optimization on two different MacBook Pro's, but I'm finding no change in speed. Is it possible that you're compiling without optimizations? This is something I would expect the compiler to take care of.

pedronahum:
Hi, 

I noticed the same when using gcc (4.8 and 6.4). Maybe is something very specific for the Intel compiler.  Will take a closer look and get back to you.

pedronahum:
To have the same "benchmark", what compiler are you using? I have a Mac that I can use for this as well.

mgbellemare:
This is with Apple LLVM 8.1 (i.e. clang).

pedronahum:
Thanks! I will profile then with TAU, which recently added supoort for Clang/LLVM, and let you know what I find.

pedronahum:
Hi,

I moved my development environment to a Mac, and I have to say I am seeing exactly the same as you. I have tested with icc 18.0.1, clang LLVM version 9.0 and pgi 17.10 and in the 3 compilers, both flattening and memory alignment didn't add any performance increase. I think we can close this pull request, as I don't think it is adding much value, in fact, the opposite, I wasted some hours of your time. Apologies for this. 

In any case, now I have a proper benchmark (i.e., not Windows with icc), and if I find some ways to vectorise some of important loops I will let you know.

mgbellemare:
No problem, thanks for looking into this! You might want to have a look at

https://github.com/deepmind/xitari

There are some optimizations in there that might be worth porting to ALE.

PR #227, opened 2017-11-27
Fix modes serialization

mgbellemare:
This should take care of serialization with modes & difficulty. Also note the change to Python interface version.


mcmachado:
LGTM. There is not much to look at regarding code.

I wrote a small code that calls `saveState`/`loadState` at every time step to let the agent always look one step ahead. The code seems to work seamless in `Space Invaders` and `Breakout`. The modes and difficulties are kept between the function calls.

PR #228, opened 2017-11-28
Fixing bug in DoubleDunk modes.

mcmachado:
The modes in DoubleDunk were wrong. This fixes it.

### DIFF
--- a/src/games/supported/DoubleDunk.cpp
+++ b/src/games/supported/DoubleDunk.cpp
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
     return modes;
 }
 
+void DoubleDunkSettings::goDown(System &system,
+                            std::unique_ptr<StellaEnvironmentWrapper> &environment) {
+    // this game has a menu that allows to define various yes/no options
+    // this function goes to the next option in the menu
+    unsigned previousSelection = readRam(&system, 0xB0);
+    while(previousSelection == readRam(&system, 0xB0)){
+        environment->act(PLAYER_A_DOWN, PLAYER_B_NOOP);
+        environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+    }
+}
+
+void DoubleDunkSettings::activateOption(System &system, unsigned byteOfInterest,
+                                    std::unique_ptr<StellaEnvironmentWrapper> &environment) {
+    // once we are at the proper option in the menu,
+    // if we want to enable it all we have to do is to go right
+    while((readRam(&system, 0x80) & byteOfInterest) != byteOfInterest) {
+        environment->act(PLAYER_A_RIGHT, PLAYER_B_NOOP);
+        environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+    }
+}
+
+void DoubleDunkSettings::deactivateOption(System &system, unsigned byteOfInterest,
+                                    std::unique_ptr<StellaEnvironmentWrapper> &environment) {
+    // once we are at the proper optio in the menu,
+    // if we want to disable it all we have to do is to go left
+    while((readRam(&system, 0x80) & byteOfInterest) == byteOfInterest) {
+        environment->act(PLAYER_A_LEFT, PLAYER_B_NOOP);
+        environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+    }
+}
+
 // set the mode of the game
 // the given mode must be one returned by the previous function
 void DoubleDunkSettings::setMode(game_mode_t m, System &system,
-                              std::unique_ptr<StellaEnvironmentWrapper> environment) {
+                                std::unique_ptr<StellaEnvironmentWrapper> environment) {
 
     if(m < getNumModes()) {
         environment->pressSelect();
+
+        bool threePts = m % 2 == 1 ? true : false;
+        bool tenSecs = m % 4 > 1 ? true : false;
+        bool threeSecs = m % 8 > 3 ? true : false;
+        bool foul = m % 16 > 7 ? true : false;
+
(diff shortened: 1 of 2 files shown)


src/games/supported/DoubleDunk.cpp
```diff
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
...
+    // this game has a menu that allows to define various yes/no options
+    // this function goes to the next option in the menu
+    unsigned previousSelection = readRam(&system, 0xB0);
+    while(previousSelection == readRam(&system, 0xB0)){
+        environment->act(PLAYER_A_DOWN, PLAYER_B_NOOP);
+        environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+    }
+}
+
+void DoubleDunkSettings::activateOption(System &system, unsigned byteOfInterest,
```
mgbellemare: unsigned int please.

src/games/supported/DoubleDunk.cpp
```diff
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
...
+                                    std::unique_ptr<StellaEnvironmentWrapper> &environment) {
+    // once we are at the proper option in the menu,
+    // if we want to enable it all we have to do is to go right
+    while((readRam(&system, 0x80) & byteOfInterest) != byteOfInterest) {
+        environment->act(PLAYER_A_RIGHT, PLAYER_B_NOOP);
+        environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+    }
+}
+
+void DoubleDunkSettings::deactivateOption(System &system, unsigned byteOfInterest,
```
mgbellemare: unsigned int please.

src/games/supported/DoubleDunk.cpp
```diff
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
...
 // set the mode of the game
 // the given mode must be one returned by the previous function
 void DoubleDunkSettings::setMode(game_mode_t m, System &system,
-                              std::unique_ptr<StellaEnvironmentWrapper> environment) {
+                                std::unique_ptr<StellaEnvironmentWrapper> environment) {
 
     if(m < getNumModes()) {
         environment->pressSelect();
+
+        bool threePts = m % 2 == 1 ? true : false;
```
mgbellemare: Simpler and clearer: m & 1, m & 2, m & 4, m & 8 (these bits are set).

In fact, you can probably do `threeSecs = (bool) m & 4;`

src/games/supported/DoubleDunk.cpp
```diff
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
...
-            }
-            environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
-            environment->act(PLAYER_A_DOWN, PLAYER_B_NOOP);
-            environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
-        }
+        goDown(system, environment);
+        goDown(system, environment);
+
+        //deal with the 3 points option
+        threePts ? activateOption(system, 0x08, environment) : deactivateOption(system, 0x08, environment);
```
mgbellemare: I think you can simplify this logic to:

`if (m & 1)
  activateOption()
else
  deactivateOption()
`

Was there a reason for avoiding this?

src/games/supported/DoubleDunk.cpp
```diff
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
...
         environment->softReset();
         //apply starting action
         environment->act(PLAYER_A_UPFIRE, PLAYER_B_NOOP);
         environment->act(PLAYER_A_NOOP, PLAYER_B_NOOP);
+        
     }
     else {
         throw std::runtime_error("This mode doesn't currently exist for this game");
     }
+    std::cout << "Byte: " << readRam(&system, 0x80) << std::endl;
```
mgbellemare: Please remove.

src/games/supported/DoubleDunk.cpp
mgbellemare: bitOfInterest, not byte. Otherwise the logic below doesn't make sense (e.g. 0x0111 & 0x1100 will give 0x0100).

src/games/supported/DoubleDunk.cpp
```diff
@@ -129,37 +129,79 @@ ModeVect DoubleDunkSettings::getAvailableModes() {
...
+        //deal with the 3 points option
+        threePts ? activateOption(system, 0x08, environment) : deactivateOption(system, 0x08, environment);
+
+        //deal with the 10 seconds option
+        goDown(system, environment);
+        tenSecs ? activateOption(system, 0x10, environment) : deactivateOption(system, 0x10, environment);
+
+        //deal with the 3 seconds option
+        goDown(system, environment);
+        threeSecs ? activateOption(system, 0x04, environment) : deactivateOption(system, 0x04, environment);
```
mgbellemare: It's a real shame these bits aren't ordered, otherwise we could have written a nice bit loop.

src/games/supported/DoubleDunk.hpp
```diff
@@ -86,6 +86,15 @@ class DoubleDunkSettings : public RomSettings {
         bool m_terminal;
         reward_t m_reward;
         reward_t m_score;
+
+        void goDown(System &system,
```
mgbellemare: Can you add a comment about these methods -- how they're used to set the mode?

mcmachado:
@mgbellemare: I changed the code accordingly. Can you take a final look?

PR #232, opened 2017-12-18
Allow setting ram

nhynes:
Being able to probe the registers would be useful, too, but this is a start.


mcmachado:
@nhynes , it is not clear to me, what is the use case for this feature?

nhynes:
It's useful when creating custom scenarios. Here's an [example](https://gist.github.com/nhynes/90fd0f2646fb1a297324e2378f25fcda) of a function that works for Pong.

mgbellemare:
Hi @nhynes , thanks for the PR -- a little busy right now but will get back to you soon as I can get a chance.

JesseFarebro:
Thanks for your contribution @nhynes. I'm closing this in favour of #377.

Issue #183, opened 2016-10-23
How can I get sound output as data just like function getScreenRGB ?


boluoweifenda:
I want to use both image and sound data to analyze agent's state,
I noticed that SDL support in ALE can output sound to mic and record a wav file, but It's arduous to extract.
Is there any way to get sound data directly from the emulator?

mgbellemare:
This isn't supported at the moment. If this is something you'd like to contribute, I'd be happy to help you get started -- email or PM me.

hholst80:
Just want to keep myself in the loop if something happens here. ALE would be a good test bed for graphics and audio input to agents.

JesseFarebro:
PR #233 introduces the ability to query the audio samples. I might have more time to refine #233 in the future but as of right now people can use @shayegano's fork with audio support.

PR #233, opened 2017-12-18
Adding audio query support

shayegano:
### Summary
Just in time for the holidays, this PR adds audio query support and addresses #183.

Here are some videos of this in action (screengrabs on left/audio spectrogram on right):
[Videos of ALE with Audio](https://www.youtube.com/watch?v=iaA8vFRIt3U&index=37&list=PLcLXhPlZoJmVSsYmfI2sHtDMMukBHja4L)

There are many games in which audio provides potentially useful observations of the latent state. For instance, audio indications for item pickups, enemy kills, collisions, crossing of boundaries, etc., are fairly standard in Atari games; such events are, sometimes, more discernible in audio than in visual inputs. We think this may enable development of interesting new algorithms that take advantage of audio as a sparse, yet useful, auxiliary percept.

### Implementation
We made an effort to keep audio query logic & API consistent with the existing codebase, and to keep any modifications simple and understandable. Some notes on this:

- Users can enable audio querying by setting flag "recording_for_user" to True. If "record_sound_filename" is also set while user queries are enabled, the latter takes precedence. Our reasoning was two-fold: 1) It would be rare to need to both query and record audio to file; 2) More importantly, it means data structures and code between the two cases can be shared, keeping things simple code for future contributors.

- Audio storage and queries are handled through the existing SoundExporter and SoundSDL classes, respectively. This was done to minimize obfuscation of the existing main game loop code, which already used SoundExporter for recording to file.

- Standard audio processing in SoundSDL::processFragment(...) is disabled when user audio queries are enabled. This enabled usage of the existing RegWriteQueue in SoundSDL to store/process the audio, for both the user query and usual playback cases. This also ensures audio is not double processed.

- Audio queries adhere to the existing frame_skip logic for screen grabs, returning only the audio for the final frame in the skipped frame batch. E.g., if frame_skip == 4, only the audio for the 4th frame is returned.

We are happy to modify any of the above per community suggestions. 

### Usage
Upon setting "recording_for_user = True", the user can call the following audio-related functions:

1) getAudio(data_audio): fills a pre-allocated data_audio array with the audio samples from the latest frame.

2) getScreenRGBAndAudio(data_screen, data_audio): fills pre-allocated arrays with image and audio data from the latest frame.

3) getAudioSize(): returns size of the current audio buffer. In the current implementation, audio buffer size is a constant 512 samples/frame, so this function is mainly a protective measure for users in case the sample rate changes in the future.

### Example Code
Two example scripts are provided:

1) Arcade-Learning-Environment/doc/examples/python_example_with_audio.py: Barebones example of querying audio.

2) Arcade-Learning-Environment/doc/examples/python_example_with_audio_advanced.py: More advanced example, which converts audio to a spectrogram image, saves synchronized screen-and-spectrogram PNGs in addition to audio WAV file, and finally uses ffmpeg to convert the captures to an MP4 movie. This is how we generated the movies linked above, and might be useful to some users. 

### Testing
Audio queries and sync with images were tested across many games, with videos recorded to verify synchronization.

Sync tests were conducted with both frame_skip 1 and 4.

Tests conducted using python_example_with_audio.py indicated very negligible performance drop due to audio queries. 

### Documentation
We will add documentation to the manual once all revisions related to this PR are complete.

### Credits
Joint credz go to my labmate Dong-Ki Kim (@dkkim93) for this contrib. Thanks Dong-Ki! 

For the curious, we did this as part of a [NIPS workshop paper](https://arxiv.org/abs/1711.10314) on crossmodal learning. Here's a gif of it in action:

SL Amidar Gameplay

This was a fun codebase to work on :)


src/emucore/Settings.cxx
```diff
@@ -234,6 +235,12 @@ void Settings::validate()
   i = getInt("tiafreq");
   if(i < 0 || i > 48000)
     setInternal("tiafreq", "31400");
+  b1 = !getString("record_sound_filename").empty();
+  b2 = getBool("record_sound_for_user");
+  //TODO(sos) ask mgb/mm: should we throw/warn user that
```
shayegano: Please double check if we need to add a user warning here.

src/environment/stella_environment.cpp
```diff
@@ -265,6 +268,14 @@ std::unique_ptr<StellaEnvironmentWrapper> StellaEnvironment::getWrapper() {
     return std::unique_ptr<StellaEnvironmentWrapper>(new StellaEnvironmentWrapper(*this));
 }
 
+const std::vector<ale::sound::SoundExporter::SampleType>& StellaEnvironment::getAudio() const {
+  if (!m_osystem->settings().getBool("sound") || !m_osystem->settings().getBool("record_sound_for_user")) {
+	  // TODO(sos) should we throw/kill? Or is this enough?
+      ale::Logger::Error << "Error: Attempted to call getAudio(), but 'sound' and/or 'record_sound_for_user' setting is false." << std::endl;
```
shayegano: Please double check if we need to be more stringent here.

src/ale_interface.hpp
```diff
@@ -143,6 +143,9 @@ class ALEInterface {
   //followed by the green colours and then the blue colours
   void getScreenRGB(std::vector<unsigned char>& output_rgb_buffer);
 
+  //This method should receive an empty vector to fill it with audio data
```
shayegano: Missed this old comment (change to "Returns the current audio data"). Will do on next commit.

mgbellemare:
Nice! This will take me a little while to parse through, but looking forward to taking a look.

melloflavio:
Hi there, I am getting myself familiarised with ALE and was wondering if was there any option to read the audio stream along with the image. Since I couldn't find anything related to that in the original docs, a google search brought me here, but this PR seems to be timed out. I was wondering if this was indeed implemented in the main ALE codebase somewhere else, or is it something still hanging?

BTW congrats to all involved in this project. It is really interesting

rienath:
This is very old and very useful! Please merge @JesseFarebro

PR #234, opened 2017-12-23
Add mode support for Galaxian

alito:


src/games/supported/Galaxian.cpp
```diff
@@ -122,3 +122,33 @@ void GalaxianSettings::loadState(Deserializer & ser) {
...
+        // press select until the correct mode is reached
+        while (mode != static_cast<unsigned>(readRam(&system, 0xB3))) {
+            environment->pressSelect();
+        }
+        //reset the environment to apply changes.
+        environment->softReset();
+    } 
+    // ignore 0 since that's the default even though it's not a 
+    // valid mode
+    else if (mode != 0)
```
mcmachado: The default mode is `0`, here you are forcing the user to set the mode, aren't you? We don't want that, we want the default mode to be used when `setMode` and `setDifficulty` is not called.

src/games/supported/Galaxian.cpp
```diff
@@ -122,3 +122,33 @@ void GalaxianSettings::loadState(Deserializer & ser) {
...
+}
+
+// set the mode of the game
+// the given mode must be one returned by the previous function
+void GalaxianSettings::setMode(game_mode_t mode, System &system,
+                              std::unique_ptr<StellaEnvironmentWrapper> environment) {
+
+    if (mode >= 1 && mode <= 9) {
+        // press select until the correct mode is reached
+        while (mode != static_cast<unsigned>(readRam(&system, 0xB3))) {
```
mcmachado: In all the other games we are not using `static_cast<unsigned>()` when reading the RAM to compare with the mode value. Why are you doing it here?

mcmachado:
Thanks @alito !

I made a couple of comments on the PR, just to clarify the code.
Aside from that, looking at the manual, Galaxian has two difficulty levels as well.

```
DIFFICULTY SWITCHES

The LEFT DIFFICULTY switch determines your laser base firing response
mode. Set the DIFFICULTY switch to B and hold the fire button down
for continuous firing. Set the DIFFICULTY switch to A and use your
thumb to tap the fire button each time you want to fire a missile.
```

Can you implement it as well? I want to make sure everything is being properly supported when adding a new game.

Cheers!

src/games/supported/Galaxian.cpp
alito: I've changed it so that if mode 0 is passed in it sets it to mode 1.

I think the patch in pull #236 is a better way of handling this though. Mode 0 doesn't really exist.

src/games/supported/Galaxian.cpp
alito: This is just to silence the (very reasonable) warning produced by gcc otherwise regarding comparing signed and unsigned integers. The cast is an explicit acknowledgement that that is what is happening. I'm happy to remove it if it's a problem. Let me know.

alito:
No problems, added and other changes implemented.

mcmachado:
LGTM.

mcmachado:
LGTM. Merging now.

PR #236, opened 2017-12-29
Let ROMs set their default mode. 

alito:
The current code sets mode 0 by default even when this mode is not in the available list of modes supported by the ROM. This leads to games like centipede and battle zone to convert mode 0 to one of their supported modes and 2 separate modes mapping to the same game.

Even in the cases where the mode is supported, this isn't the real name of the mode, it is just a mapped name from its "real" number to 0, introducing an unnecessary level of semantic indirection.

This patch lets ROMs set their default mode. It also sets a default implementation which returns the first mode in the list of available modes. It also should lead to more meaningful use of mode numbers in the future.

I've checked every rom and this leads to a non-zero implementation in the following cases:

air_raid: 1
battle_zone: 1
berzerk: 1
centipede: 22
defender: 1
demon_attack: 1
name_this_game: 8
pooyan: 10

In all these cases, this is the mode that gets set when passed 0 in any case, so there are no changes in behaviour due to this patch.

### DIFF
--- a/src/environment/stella_environment.cpp
+++ b/src/environment/stella_environment.cpp
@@ -38,7 +38,10 @@ StellaEnvironment::StellaEnvironment(OSystem* osystem, RomSettings* settings):
   }
   m_num_reset_steps = 4;
   m_cartridge_md5 = m_osystem->console().properties().get(Cartridge_MD5);
-  
+ 
+  // Set current mode to the ROM's default mode
+  m_state.setCurrentMode(settings->getDefaultMode());
+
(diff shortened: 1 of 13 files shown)


mcmachado:
I like the idea. @mgbellemare , should this be pushed to the `dev` or to the `master` branch?
@alito for this PR to be complete, you have to get rid of the `if` statements in each one of the games you listed above, don't you? For example, in `Centipede`, the if-statement:

```
if (m == 0) {
        m = 0x16; // The default mode doesn't work here.
    }
```
shouldn't be necessary anymore, right?

alito:
Sorry, I mangled the rebase on this. Creating a new pull request

PR #237, opened 2017-12-30
Let ROMs set their default mode. 

alito:
(Copy of #236 due to rebase mangling on my part. Also now removes the check for whether the mode equaled zero in all the games that did not have a zero mode listed)

The current code sets mode 0 by default even when this mode is not in the available list of modes supported by the ROM. This leads to games like centipede and battle zone to convert mode 0 to one of their supported modes and 2 separate modes mapping to the same game.

Even in the cases where the mode is supported, this isn't the real name of the mode, it is just a mapped name from its "real" number to 0, introducing an unnecessary level of semantic indirection.

This patch lets ROMs set their default mode. It also sets a default implementation which returns the first mode in the list of available modes. It also should lead to more meaningful use of mode numbers in the future.

I've checked every rom and this leads to a non-zero implementation in the following cases:

air_raid: 1
battle_zone: 1
berzerk: 1
centipede: 22
defender: 1
demon_attack: 1
name_this_game: 8
pooyan: 10

In all these cases, this is the mode that gets set when passed 0 in any case, so there are no changes in behaviour due to this patch. I've removed the corresponding checks for zero in all of these games

### DIFF
--- a/src/environment/stella_environment.cpp
+++ b/src/environment/stella_environment.cpp
@@ -46,6 +46,9 @@ StellaEnvironment::StellaEnvironment(OSystem* osystem, RomSettings* settings)
   m_num_reset_steps = 4;
   m_cartridge_md5 = m_osystem->console().properties().get(Cartridge_MD5);
 
+  // Set current mode to the ROM's default mode
+  m_state.setCurrentMode(settings->getDefaultMode());
+
   m_max_num_frames_per_episode =
       m_osystem->settings().getInt("max_num_frames_per_episode");
   m_colour_averaging = m_osystem->settings().getBool("color_averaging");
--- a/src/games/RomSettings.cpp
+++ b/src/games/RomSettings.cpp
@@ -63,6 +63,16 @@ void RomSettings::setMode(game_mode_t m, System&, std::unique_ptr<StellaEnvironm
   }
 }
 
+game_mode_t RomSettings::getDefaultMode() {
+  // By default, return the first available mode, or 0 if none are listed
+  ModeVect available_modes = getAvailableModes();
+  if (available_modes.empty()) {
+    return 0;
+  } else {
+    return available_modes[0];
+  }
+}
+
 DifficultyVect RomSettings::getAvailableDifficulties() {
   return DifficultyVect(1, 0);
 }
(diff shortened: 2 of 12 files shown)


mcmachado:
LGTM. However, shouldn't we first merge this PR into the `dev` branch?
What do you say, @mgbellemare ?

mgbellemare:
Hi @alito , apologies for delays -- will look into this soon.

kieranhj:
@mgbellemare Hi Marc, this looks like a sensible set of changes to me. It is confusing that game mode 0 is always applied to the ROM settings when the environment is created, even if it is not supported. WDYT?

justinkterry:
@mgbellemare @mcmachado Should this be merged?

tkoeppe:
@kieranhj: Now that we have ROM-overridable settings, does that somehow give us a better solution?

tkoeppe:
@alito: could you kindly rebase this PR so it integrates cleanly?

@mgbellemare: This change would be helpful and would make ALE more expressive, since it would make the default-mode choice explicit and queryable, rather than being a secret implementation detail that we nonetheless need to rely on.

mgbellemare:
Hi, somehow this PR keeps failing to make it on my radar. It seems like a sensible improvement.

alito:
I'll rebase this coming weekend. (My initial description of this patch sounds unintelligible to me by now, sorry about that)

justinkterry:
@alito ?

alito:
Rebased but untested since the Python setup is currently broken and cmake is not yet my friend. 

The recently added VideoCheckers, SpaceWar, MarioBros, FlagCapture and BasicMath have hacks around mode-setting which could be removed with this patch but I'm not sure if the hacked modes are being used by callers to the library so I left them as they were.

justinkterry:
@tkoeppe any thoughts on the mode setting at issue in those games? I believe that's your code.

kieranhj:
Support for those new ROMs was actually my code, @tkoeppe is kindly maintaining the upstream integration for us. I know we have folks using these ROMs internally so suggest leaving things as-is for now. I will take another look whether the 'start from 0' hacks can be removed when we take the changes.

justinkterry:
@kieranhj Anything new?

justinkterry:
@kieranhj Just a friendly reminder that this exists

tkoeppe:
Thanks, @kieranhj! I'd be happy to land this PR as it is then, and we can provide the updates for the ROMs to undo the mode hacks as a follow-up. @mgbellemare: PTAL?

mgbellemare:
LGTM!

PR #238, opened 2018-01-01
Do not update screen or RAM state unless it's asked for

alito:
In the usual case where the client-side is doing "manual" frameskipping, due to the max-of-last-two-frames use (see https://github.com/openai/baselines or @spragunr code), only two out of every four frames need to be generated. This patch avoids generating this, and/or the RAM state, for those frames which are not being requested. When measured on a closed loop, this raises the frames per second about 2 percent (which is a bit disappointing, yes). It does save the computer from moving around about 600 MB/s, so maybe I can justify it on ecological reasons :) Or maybe it can become important if running tons of parallel environments and the computer becomes memory-bandwidth-bottlenecked at some point. In any case, it seems like the right thing to do, with very little complexity increase.

The patch also smuggles in a small fix so that the screen recorder doesn't record duplicate frames at the end of episodes when frameskip is bigger than 1 (and some potential sound issues) by only saving when an emulation step is performed.

### DIFF
--- a/src/environment/stella_environment.cpp
+++ b/src/environment/stella_environment.cpp
@@ -25,6 +25,8 @@ StellaEnvironment::StellaEnvironment(OSystem* osystem, RomSettings* settings):
   m_phosphor_blend(osystem),  
   m_screen(m_osystem->console().mediaSource().height(),
         m_osystem->console().mediaSource().width()),
+  m_screen_updated(false),
+  m_ram_updated(false),
   m_player_a_action(PLAYER_A_NOOP),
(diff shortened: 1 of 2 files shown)


mcmachado:
@alito , the continuous integration failed for this PR.
I don't know if you want to wait,  to hear from @mgbellemare, about whether he would be interested in this PR, before fixing it, but right now it seems to be broken.

alito:
@mcmachado from what I can see in the details it seems to be a CI environment issue, not a code issue:

    oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"

PR #247, opened 2018-06-26
Add ```setRAM()``` function for python interface.

YuhangSong:
I am trying to find which part of the RAM controls the stochasticity of Atari, thus I added this ```setRAM()``` function for python and have verified that it works as expected.

I open this pull request in case anyone need the similar feature.


ale_python_interface/ale_python_interface.py
```diff
@@ -261,6 +263,15 @@ def getRAM(self, ram=None):
         ale_lib.getRAM(self.obj, as_ctypes(ram))
         return ram
 
+    def setRAM(self, ram=None):
+        """This function set the atari RAM.
+        ram MUST be a numpy array of uint8/int8. This can be initialized like so:
+        ram = np.array(ram_size, dtype=uint8)
+        Notice: It must be ram_size where ram_size can be retrieved via the getRAMSize function.
+        If it is None,  then this function will initialize it.
```
mgbellemare: Does ram=None work?

ale_python_interface/ale_c_wrapper.h
```diff
@@ -63,6 +63,12 @@ extern "C" {
     int size = ale->getRAM().size();
     memcpy(ram,ale_ram,size*sizeof(unsigned char));
   }
+  void setRAM(ALEInterface *ale,unsigned char *ram){
```
mgbellemare: This seems to satisfy a very specific use case, where you might want to set the whole RAM state but nothing else. I'm not sure what usefulness that has. Can you explain a little more what you have in mind? I would imagine setRAM(ale, memory_index, value) is a much more useful function.

src/environment/stella_environment.cpp
```diff
@@ -271,15 +271,20 @@ void StellaEnvironment::processScreen() {
...
 }
 
 void StellaEnvironment::processRAM() {
   // Copy RAM over
   for (size_t i = 0; i < m_ram.size(); i++)
-    *m_ram.byte(i) = m_osystem->console().system().peek(i + 0x80); 
+    *m_ram.byte(i) = m_osystem->console().system().peek(i + 0x80);
 }
 
+void StellaEnvironment::processBackRAM() {
```
mgbellemare: This feels like the wrong interface. Why would you have two separate methods -- one for setting the RAM as read by the ALE, and another one to then copy it back to the emulator? The alternative (set the emulator RAM directly, mirror in ALE) seems much more natural.

JesseFarebro:
Thanks for your contribution @YuhangSong. I'm closing this in favour of #377.

PR #265, opened 2019-07-09
Optimize Atari emulator by ~30%

qstanczyk:
Eliminate variables which are constant.
Inline memory access methods.
Benchmarking shows avg. ~30% speedup on multiple Atari games.

### DIFF
--- a/src/emucore/Console.cxx
+++ b/src/emucore/Console.cxx
@@ -148,7 +148,7 @@ Console::Console(OSystem* osystem, Cartridge* cart, const Properties& props)
   mySwitches = new Switches(*myEvent, myProperties);
 
   // Now, we can construct the system and components
-  mySystem = new System(13, 6);
+  mySystem = new System();
 
   // Inform the controllers about the system
   myControllers[0]->setSystem(mySystem);
--- a/src/emucore/m6502/src/System.cxx
+++ b/src/emucore/m6502/src/System.cxx
@@ -28,20 +28,13 @@
 using namespace std;
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-System::System(uInt16 n, uInt16 m)
-  : myAddressMask((1 << n) - 1),
-    myPageShift(m),
-    myPageMask((1 << m) - 1),
-    myNumberOfPages(1 << (n - m)),
-    myNumberOfDevices(0),
+System::System()
+  : myNumberOfDevices(0),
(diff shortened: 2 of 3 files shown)


mgbellemare:
LGTM, thanks for this big performance improvement! I'm going to wait for @mcmachado to give his thumbs up before merging (==> he will merge).

maxwellreuter:
@qstanczyk I'm looking to emulate my Atari games as fast as possible. Do you have a feeling for how much room there is for optimization, if any? If so, what sort of speedup % do you think is possible?

qstanczyk:
Hi Max, I didn't have time myself to do that, but I had an idea of providing games' ROM converter to C++. Currently ALE implements full Atari instruction set and performs interpretation of ROM. Instead ROM could be converted to C++/intermediate representation and JIT-compiled. That could give order of magnitude speedup, but is a longer effort.

maxwellreuter:
@qstanczyk thanks for your reply. Do you know how much of an effort that would be and what sort of skills/tools would be needed to do that?

qstanczyk:
It requires C++ knowledge and some profiling/optimization experience. Would be best I guess if instead of generating C++ it would generate Intermediate Representation and JIT with LLVM (https://llvm.org/docs/LangRef.html). This approach is basically a re-write of ALE.

maxwellreuter:
@qstanczyk ok, thanks -- do you know what sort of time commitment that would be? would it be less if I only focused on one game such as breakout? If so how much of a time commitment would that be?

qstanczyk:
With some LLVM prior experience... I would say 3 weeks of actual work. Doing that for one specific game might be slightly easier, as ROMs have different versions with different operation sets (so doing that as a proof of concept for some specific game is a good start anyway).

maxwellreuter:
@qstanczyk how long do you think it would take for someone to get up to speed without any experience with LLVM, working with ROMs, or profiling/optimization, so that they can be effective?

JesseFarebro:
@maxwellreuter if you haven't heard of the [CUDA port of the ALE](https://github.com/NVlabs/cule) you could also check that out. Depending on what you're trying to do that could provide some wall-time improvements.

qstanczyk:
Hard to tell, depends on the background, approach (do you want to understand details or just get things done)... but Jesse has a point - GPU environment might be a fast option for you.

maxwellreuter:
@JesseFarebro @qstanczyk thanks!

PR #270, opened 2019-09-25
Python interface supports arbitrary string types

### LINKED ISSUE #269: wrong string types passed to setBool in python_example.py
In python_example.py, when setting USE_SDL to True, a few parameters are set in ale. The strings for the parameternames which are passed to these setBool calls are normal python strings, instead of binary strings. This causes an ArgumentError.
In other words:
`ale.setBool('sound', True)` --> `ale.setBool(b'sound', True)`

alcinos:
Fixes #269, and more generally the interface when using python3


justinkterry:
Shouldn't this be closed in light of PR #272 ?

JesseFarebro:
Fixed in #272.

PR #271, opened 2019-09-25
atari-py 1/3: Windows build

JesseFarebro:
This is the first PR in a set of PRs merging [openai/atari-py](https://github.com/openai/atari-py) into the ALE. This first PR adds the ability to build the ALE on Windows.

To get a proper build on Windows we had to:

1. Add zlib to the repository. This is how Stella now deals with zlib as well ([stella/src/zlib](https://github.com/stella-emu/stella/tree/master/src/zlib)).
2. CMake changes: Hide some flags under a GCC check, export all Windows symbols, and link zlib properly.

For all these PRs I'm hoping @christopherhesse will be able to review them before @mgbellemare takes a look.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,9 +10,17 @@ option(BUILD_CPP_LIB "Build C++ Shared Library" ON)
 option(BUILD_CLI "Build ALE Command Line Interface" ON)
 option(BUILD_C_LIB "Build ALE C Library (needed for Python interface)" ON)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wunused -fPIC -O3 -fomit-frame-pointer -D__STDC_CONSTANT_MACROS")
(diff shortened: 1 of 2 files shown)


mgbellemare:
Hi Jesse, just coming to this now, sorry for the very long delay. Is this PR still good to go? There is a conflict that needs resolving (you can see there are other PRs coming in re: cleaning whitespace).

mgbellemare:
The inclusion of zlib is less than desirable for me -- I feel like we stripped it out at some point. Do you know if it would be possible to only bind with external/zlib when compiling under windows?

JesseFarebro:
Hey Marc,

My original reasons to include zlib were to make it easier for Windows users to build the ALE. I think this was misguided because ultimately I just need to make Windows builds easier for CI/CD. Most users will be installing the prebuilt Python package instead of building themselves.

I refactored this PR to remove zlib as I now have a clean solution for installing zlib on Windows for CI/CD which will be in the 3rd and final PR in this series.

Now, this PR just includes `CMakeLists.txt` changes for Windows and a missing include which was discovered while debugging the Windows builds.

PR #272, opened 2019-09-26
atari-py 2/3: Python interface

JesseFarebro:
This PR merges the core changes from [openai/atari-py](https://github.com/openai/atari-py). There are no major changes to the API other than the fix for #269 which had been in `atari-py` for a while.

We also include a test to make sure building the library and packaging the wheel is successful. We load `tetris.bin` which is a freely available game developed by Colin Hughes and can be found here: [https://atariage.com/software_page.php?SoftwareID=3290](https://atariage.com/software_page.php?SoftwareID=3290).

The last change is to `setup.py`. I largely reworked the CMake extension from `atari-py`. It makes packaging and building on all platforms easier. I also added more metadata for PyPi.

The last PR will be modifications for Travis to build and publish the Python wheels. In doing so we need a package name. @christopherhesse had suggested we could use `ale-py` (hence the rename in this PR). Thoughts?


setup.py
```diff
@@ -1,25 +1,96 @@
...
+                "-DBUILD_EXAMPLES=OFF",
+                "-DBUILD_CPP_LIB=OFF",
+                "-DBUILD_CLI=OFF",
+                "-DBUILD_C_LIB=ON",
+            ],
+        )
+    ],
+    cmdclass={"build_ext": CMakeBuild},
+    packages=["ale_py"],
+    install_requires=["numpy", "six"],
```
christopherhesse: Why is `six` required?

christopherhesse:
Also I see the test was added, but I don't see it being run in travis.  Are the tests to be enabled later?

setup.py
```diff
@@ -1,25 +1,96 @@
...
+    keywords=["reinforcement-learning", "arcade-learning-environment", "atari"],
+    url="https://github.com/mgbellemare/Arcade-Learning-Environment",
+    author="Arcade Learning Environment Authors",
+    license="GPL",
+    ext_modules=[
+        CMakeExtension(
+            "ale_py.libale_c",
+            ".",
+            [
+                "-DUSE_SDL=OFF",
```
christopherhesse: Aren't some of these the same as the defaults?  Might make sense not to include them if they are the default.

setup.py
```diff
@@ -1,25 +1,96 @@
...
+                ["cmake", "--build", "."] + cmake_build_args, cwd=self.build_temp
+            )
+
+
+with open("README.md", "r") as f:
+    long_description = f.read()
+
+setup(
+    name="ale-py",
+    version="0.6.0",
```
christopherhesse: Should this be a new version number?

setup.py
```diff
@@ -1,25 +1,96 @@
...
+
+            subprocess.check_call(
+                ["cmake", ext.sourcedir] + cmake_config_args, cwd=self.build_temp
+            )
+            subprocess.check_call(
+                ["cmake", "--build", "."] + cmake_build_args, cwd=self.build_temp
+            )
+
+
+with open("README.md", "r") as f:
```
christopherhesse: Does this work if you do `pip install -e Arcade-Learning-Environment`? What about `python Arcade-Learning-Environment/setup.py develop`?  I noticed we do relative paths here: https://github.com/openai/atari-py/blob/master/setup.py#L7

setup.py
```diff
@@ -1,25 +1,96 @@
...
+            # -DCMAKE_BUILD_TYPE doesn't work on Windows
+            # we need to specify --config Release at build type
+            if platform.system() == "Windows":
+                cmake_build_args += ["--config", cfg]
+
+            if not os.path.exists(self.build_temp):
+                os.makedirs(self.build_temp)
+
+            subprocess.check_call(
+                ["cmake", ext.sourcedir] + cmake_config_args, cwd=self.build_temp
```
christopherhesse: You should make sure you can do `pip install`, `pip install -e` and `python setup.py (install|develop)`, I remember having some issues where like only some of those would work so we had some crazy hack here: https://github.com/openai/atari-py/blob/master/setup.py#L17

christopherhesse: Such as https://github.com/openai/atari-py/issues/44 though I don't remember what that particular line fixes.

setup.py
JesseFarebro: It's not. Good catch!

setup.py
JesseFarebro: I would kind of like to keep these. I can't see the defaults changing but this way it's expressive that the Python wrapper only requires the C library.

setup.py
JesseFarebro: Yes, I still have to figure out a better way to manage versions. Right now I do deployments on tagging. You had mentioned this might cause some issues.

setup.py
JesseFarebro: Ahh yes, this will be a problem for reading the README.

setup.py
JesseFarebro: All the different combinations of `pip install (-e) [ . | Arcade-Learning-Environment ]` work fine. I don't think `python Arcade-Learning-Environment/setup.py (develop|install)` should work? It won't be able to find the package data?

I did find one issue though. When doing develop builds with the editable flag it seems to expect the built library to have a platform-specific suffix. This seems to be new in Python 3, I'm not sure why this wasn't affecting `atari-py`. I have to resort to getting the proper suffix from `distutils.sysconfig` when building and at runtime.

JesseFarebro:
> Also I see the test was added, but I don't see it being run in travis. Are the tests to be enabled later?

Yes, the tests are run by the new Travis CI scripts that will be in the third PR.

I also modified the build process to work with this Python `EXT_SUFFIX` (see comment above) that was giving me problems. This mainly consisted of passing a custom CMake variable if the suffix exists and making library loading more robust.

setup.py
christopherhesse: I think you suggested a good check to avoid forgetting to update the number.  The only remaining downside is that if you tag a release before tests pass, you may have to re-tag it.  You can just wait until tests pass before tagging though.

setup.py
christopherhesse: Yeah, I think `setup.py` has to be called when it's in the CWD.

setup.py
mgbellemare: Should this be a new version number?

JesseFarebro: I have some additional work on versioning that I plan on pushing to this PR. The plan is to extract the version from `CMakeLists.txt` so we can have one source of truth for the version info.

JesseFarebro: Versioning `ale-py` works as follows:

The version should be bumped in `CMakeLists.txt`. `setup.py` will parse out the version and append a suffix of `dev` if we aren't running in CI. If we are running in CI and the tagged version matches the parsed version from `CMakeLists.txt` the suffix will be removed and setup will continue.

There is also a check that both the parsed and tagged versions conform to semver rules. This is mainly due to PyPi being picky so we might as well catch it before the rest of the deployment automation continues.

JesseFarebro:
This looks all good to me now. I already have full continuous deployment working for `ale-py` with these new versioning changes.

I'll explain the journey of deployment in the third PR.

ale_py/ale_python_interface.py
```diff
@@ -2,15 +2,46 @@
...
+    if sys.platform.startswith("linux"):
+        ext_suffix = ".so"
+    elif sys.platform.startswith("darwin"):
+        ext_suffix = ".dylib"
+    elif sys.platform.startswith("win"):
+        ext_suffix = ".dll"
+    else:
+        raise RuntimeError(
+            'Platform "%s" not recognized while trying to resolve shared library'
+            % sys.platform
```
mgbellemare: This is a place where Google Python style seems to disagree with PEP-8, but: can you move the closing parentheses to be on the same line as the last part of their statement? The cognitive dissonance is too strong. See http://google.github.io/styleguide/pyguide.html

JesseFarebro: I changed both of the exceptions to use `format` for consistency and fixed the parentheses.

mgbellemare:
@JesseFarebro , what's the status of this PR again?

JesseFarebro:
It should be good to merge. I addressed the formatting issues in your last review. You can also close #270 and #149 , this PR resolves those issues with the Python wrapper.

PR #274, opened 2019-11-19
Reformat non-games ALE files

tkoeppe:
This change only modifies whitespace (and adds one or two #ifdef
comments). The reformatting is based on clang-format, with manual
review and tweaking.

The goal to harmonize the source code formatting, which is currently
using a variety of local styles. The chosen format is a "compact"
one (as per clang-format default): 2-space indent, Egyptian braces.
Nonetheless, vertical whitespace has been adjusted generously, so as
to have line comments preceded by a blank line for easy visual
segmentation.

### DIFF
--- a/doc/examples/fifoInterfaceExample.cpp
+++ b/doc/examples/fifoInterfaceExample.cpp
@@ -24,118 +24,112 @@
 #include <cstdlib>
 
 // From RL-Glue agent example.
-int randInRange(int max){
-	double r, x;
-	r = ((double)rand() / ((double)(RAND_MAX)+(double)(1)));
-   	x = (r * (max+1));
-	return (int)x;
+int randInRange(int max) {
+  double r, x;
+  r = ((double)rand() / ((double)(RAND_MAX) + (double)(1)));
+  x = (r * (max + 1));
+  return (int)x;
 }
-
 
 // Print the RAM string
 void printRAM(char* str) {
+  // First we parse the ram (for pedagogical purposes)
+  std::vector<int> ram;
 
-    // First we parse the ram (for pedagogical purposes)
-    std::vector<int> ram;
+  for (int offset = 0; offset < 128; offset++) {
+    // Crude but effective
+    char buffer[16];
+    buffer[0] = str[offset * 2];
+    buffer[1] = str[offset * 2 + 1];
+    buffer[2] = 0;
 
-    for (int offset = 0; offset < 128; offset++) {
+    int value = strtol(buffer, NULL, 16);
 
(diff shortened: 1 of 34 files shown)


doc/examples/fifoInterfaceExample.cpp
```diff
@@ -1,141 +1,133 @@
 /* *****************************************************************************
  * A.L.E (Arcade Learning Environment)
  * Copyright (c) 2009-2013 by Yavar Naddaf, Joel Veness, Marc G. Bellemare,
- *  Matthew Hausknecht, and the Reinforcement Learning and Artificial Intelligence 
- *  Laboratory
- * Released under the GNU General Public License; see License.txt for details. 
+ *  Matthew Hausknecht, and the Reinforcement Learning and Artificial
+ *Intelligence Laboratory Released under the GNU General Public License; see
```
mgbellemare: Is the removal of the newline intentional here? the first line lists the ALE authors, the second is about the license.

tkoeppe: That's totally bogus! Hang on, I'll review everything.

tkoeppe:
#277 will simplify this change quite a bit

tkoeppe:
Hold off on this, I think I need to reconstruct this change.

tkoeppe:
Another thing that was bogus was that header inclusions were reordered, which is broken because the code isn't modular.

In my new approach I use a more conservative formatting setting, namely:

```
clang-format --sort-includes=false --style='{ReflowComments: false, PointerAlignment: Left}'
```

src/common/Constants.cpp
```diff
@@ -16,51 +16,57 @@
...
+      "PLAYER_B_DOWNLEFT",
+      "PLAYER_B_UPFIRE",
+      "PLAYER_B_RIGHTFIRE",
+      "PLAYER_B_LEFTFIRE",
+      "PLAYER_B_DOWNFIRE",
+      "PLAYER_B_UPRIGHTFIRE",
+      "PLAYER_B_UPLEFTFIRE",
+      "PLAYER_B_DOWNRIGHTFIRE",
+      "PLAYER_B_DOWNLEFTFIRE",
+      "__invalid__" // 36
```
mgbellemare: It looks like something went wrong here, please fix.

tkoeppe: Yes, thanks! I had previously fixed this, but when I restarted this change, I forgot. I'll do another full review pass, stay tuned.

tkoeppe: OK, that's done now, PTAL!

PR #275, opened 2019-11-19
Change line endings to LF

tkoeppe:
This affects only a few odd files that had CRLF endings (82 out of 435); most files already have LF endings.

(If needed, line ending translation should be done locally by git, and not checked in.)


tkoeppe:
@mgbellemare: Please merge #274 first. I'm not sure how easy it will be to rebase these EOL changes, and this PR is much easier to recreate than the other.

tkoeppe:
Actually, please do let's do this change first, and I can make the other changes more granular.

PR #278, opened 2019-11-22
[travis] Adds a check for trailing whitespace.

tkoeppe:


tkoeppe:
(This change requires that we first fix all the trailing whitespace, of course.)

tkoeppe:
The weirdest thing happened: `find -exec` seems to have truncated the match list, and results in fewer executions that `find | xargs`! That's why lots of files were missed. See #279 for the remaining cases.

PR #287, opened 2019-12-09
[README] Reformat markdown sources

tkoeppe:
This allows easier editing and maintenance of the README file.


tkoeppe:
I also added an "Overview" second-level heading to avoid hanging paragraphs.

PR #293, opened 2020-01-15
Move ALE C++ code into namespace "ale"

tkoeppe:


tkoeppe:
In case anyone is affected or broken by this change, please note that you can add a using-directive such as `using namespace ::ale;` somewhere near the top of your dependencies to continue being able to find ale names in the global namespace.

Issue #292, opened 2020-01-11
segmentation fault in ale.loadROM


AbhinavGopal:
When I attempt to run doc/example/ale-sharedLibraryInterfaceExample SpaceInvaders.bin, I receive the following segfault which occurs during ale.loadROM.

doc/examples/ale-sharedLibraryInterfaceExample Space\ Invaders.bin 
A.L.E: Arcade Learning Environment (version 0.6.0)
[Powered by Stella]
Use -help for help screen.
Game console created:
  ROM file:  Space Invaders.bin
  Cart Name: Space Invaders (1978) (Atari) [!]
  Cart MD5:  72ffbef6504b75e69ee1045af9075f66
  Display Format:  AUTO-DETECT ==> NTSC
  ROM Size:        4096
  Bankswitch Type: AUTO-DETECT ==> 4K

Screen Display Active. [Manual Control Mode] 'm' [Slowdown] 'a' [Speedup] 's' [VolumeDown] '[' [VolumeUp] ']'.
Running ROM file...
Random seed is 123
Segmentation fault: 11

I also receive the following error message (not sure how useful this is). Am I missing a step before this? I have already run the make and pip install steps.

Process:               ale-sharedLibraryInterfaceExample [3404]
Path:                  /Users/USER/Desktop/*/ale-sharedLibraryInterfaceExample
Identifier:            ale-sharedLibraryInterfaceExample
Version:               0
Code Type:             X86-64 (Native)
Parent Process:        bash [1045]
Responsible:           ale-sharedLibraryInterfaceExample [3404]
User ID:               501

Date/Time:             2020-01-11 14:09:49.577 -0800
OS Version:            Mac OS X 10.14.6 (18G2022)
Report Version:        12
Bridge OS Version:     4.2 (17P2551)
Anonymous UUID:        48506A87-20BD-9B7A-47DF-8FFD56884475

Sleep/Wake UUID:       99CC8CDD-6B7C-43E9-B3EC-4E65FB396041

Time Awake Since Boot: 84000 seconds
Time Since Wake:       740 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

VM Regions Near 0:
--> 
[...]

PR #294, opened 2020-01-15
User feedback if we can't find a valid wrapper

JesseFarebro:
Give useful feedback if we can't valid a valid ROM wrapper. Hopefully, this will prevent issues like #292.

I just refactored `checkForUnsupportedRom` into `isSupportedRom` so we can give some useful feedback if the ROM isn't supported and we failed to find a wrapper.

### DIFF
--- a/src/ale_interface.cpp
+++ b/src/ale_interface.cpp
@@ -82,7 +82,7 @@ void ALEInterface::createOSystem(std::unique_ptr<OSystem>& theOSystem,
   theOSystem->settings().loadConfig();
 }
 
-void ALEInterface::checkForUnsupportedRom(
+bool ALEInterface::isSupportedRom(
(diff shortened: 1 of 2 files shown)


tkoeppe:
Just to confirm, is a lack of MD5 validation still tolerated? (I depend on this, and it would be very awkward for me to get MD5 validation to work, on account how unconfigurably hardcoded the dependency on the current working directory is.)

JesseFarebro:
Hi @tkoeppe,

Yes, mismatched MD5s are still valid. All this does is give a useful warning for:

1) The ROM is known to be supported so we warn about filename conventions.
2) Warn that the ROM may be unsupported and warn about filename conventions.

As long as the ROM adheres to the filename convention it will build the corresponding wrapper.

tkoeppe:
Thanks a lot, @JesseFarebro!

PR #298, opened 2020-01-23
[emucore/TIA.cxx] Make initialization of static data thread-safe.

tkoeppe:
The initializing functions are now guarded with std::call_once.

It is someone annoying that we are computing these values at runtime
even though they are perfectly well deterimined statically. However,
the programmatic expressions may be more meaningful than the raw
numbers, so it is useful to retain the code. We _could_ make the code
evaluate as constant expressions, but unfortunately this seems to
require non-standard toolchain settings to allow for a rather large
constant evaluation budget during compilation.

The upstream Stella code has diverged from this code anyway, so this
is just a local fix for the time being.

### DIFF
--- a/src/emucore/TIA.cxx
+++ b/src/emucore/TIA.cxx
@@ -20,6 +20,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <iostream>
+#include <mutex>
 
 #include "Console.hxx"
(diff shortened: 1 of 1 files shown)


mgbellemare:
This is starting to be deep into the Stella code! Thanks for the fix.

PR #303, opened 2020-02-03
[games/DoubleDunk] Fix signedness

tkoeppe:


mgbellemare:
This sounds about right, but let me confirm with @JesseFarebro who wrote this code. @JesseFarebro , was there a reason for the unsigned int?

JesseFarebro:
@mgbellemare did you mean @mcmachado? It seems he wrote this code in this commit:

https://github.com/mgbellemare/Arcade-Learning-Environment/blob/f0dda99c60f4ac2b7ecea0bdfbf39c9d703f2aa4/src/games/supported/DoubleDunk.cpp#L136

It was discussed in this PR from the looks of it: https://github.com/mgbellemare/Arcade-Learning-Environment/pull/228.

tkoeppe:
(Interesting. The review discussion there seemed to be mainly about "unsigned" vs "unsigned int" (i.e. a mere style question. The code change here doesn't do any bitwise operations, and `readRam` returs a signed int, so it seems awkward to convert it there and back again for no clear benefit.

PR #305, opened 2020-02-07
ale-py 3/3: Deployment

JesseFarebro:
This PR has been simmering for a while. I wanted a solution to deploying ale-py which would be as hands-off as possible. It's not the easiest thing to build wheels for every possible Python version and platform.

## Why not Travis?
In the beginning, I was trying to use Travis CI but there were a couple of issues.

1. There's no easy way to store artifacts between tasks. This means you either try to deploy a wheel to PyPi before you know if all the builds succeeded OR you store the wheels on a cloud provider like AWS and deploy after you know all the builds finished successfully. Neither of these solutions seemed acceptable to me.
2. Windows builds are painful. There are a lot of small issues with Windows builds on Travis. There's a bug with secrets being leaked on PyPi deployment, it's also hard to configure and install zlib and SDL.
3. There was no clear solution to managing deployments, you either automate deployment completely or you manually deploy all the wheels yourself. I wanted a solution where we could create a release and stage the wheels to PyPi Test and then deploy them once we're happy with the release.

## Deployment with Github Actions

Here is the workflow with Github actions.

1. The version is bumped in `CMakeLists.txt`. A tag is created for the version.
2. Python 3+ wheels are built for Windows x64, macOS, and Linux x64.
3. A __DRAFT__ release is created automatically with the tag and all the wheels are listed as an artifact.
4. The wheels are uploaded to PyPi Test, a staging test server for PyPi.
5. Once we are happy with the __DRAFT__ release and you click "Publish Release" then the wheels are uploaded to the public PyPi server.

The only manual intervention for releases are:

1. Bump version in `CMakeLists.txt`
2. Tag the repository `git tag v0.6.2 && git push origin --tags`
3. Write a release summary and click "Publish Release"

This is a much cleaner solution and allows for a staging process before publishing to PyPi. This is important because you can't replace PyPi wheels, once a version string is used you can't reuse it.

## CI Builds

I've also created an action that performs builds of the ALE on any commit or PR. This now builds for each platform (macOS, Windows, and Linux) and also integrates the changes from @tkoeppe and his presubmit script.

Thanks to @christopherhesse for answering all my questions on the pain points of their deployment. If you want to see these actions being run you can check out the master branch on my fork: [JesseFarebro/Arcade-Learning-Environment](https://github.com/JesseFarebro/Arcade-Learning-Environment).


.github/workflows/build-ale-py.yml
```diff
@@ -0,0 +1,83 @@
...
+        os: [ubuntu-latest, windows-latest, macos-latest]
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      name: Install Python ${{ matrix.python_version }}
+      with:
+        python-version: "3.7"
+    - name: Install dependencies
+      run: |
+        python -m pip install git+https://github.com/mayeut/cibuildwheel.git@github-workflows
```
JesseFarebro: We need to wait until there is an official release of `cibuildwheel`. This is currently their Github actions branch. It seems like it should be happening in the next couple weeks.

src/common/display_screen.h
```diff
@@ -17,6 +17,8 @@
 #ifndef DISPLAY_SCREEN_H
 #define DISPLAY_SCREEN_H
 
+#include <algorithm>
```
JesseFarebro: It seems Windows defines `min` as `__min` in `stdlib.h`. Instead of using ifdef guards on the platform just use `min` from `algorithm.h`. We already link against `algorithm`.

christopherhesse:
Looks great, I used travis for the [procgen](https://github.com/openai/procgen) builds and ran into all the issues you mentioned, github actions sounds like it could be pretty good.

Does your setup mean that if, say, 0.7.1 is pushed to pypi test but sucks (has a bug or whatever), you have to increment the version number to 0.7.2 in order to try again?

JesseFarebro:
@christopherhesse Right now I would need to make a different version manually like `v0.7.1-dev(X+1)`. At one point I wanted to tag the PyPi test release with a short commit SHA but then I found out you can't use any tag which can't be sorted lexicographically.

I was thinking of using the # of commits since the last tag as a suffix, e.g., `0.7.1-devX` where X is the number of commits since the last tag. This wasn't a great solution either because of being able to force push and it gets messy when dealing with merging branches.

It might be possible to use `0.7.1.devDATE` where DATE is like `20200207`. I would need to add some other unique identifier like the time.

Perhaps one of these solutions could work, I just thought this would happen infrequently.

JesseFarebro:
@christopherhesse I found a solution, Github Actions provides a build number which we can append to the PyPi test wheels.

https://github.com/mgbellemare/Arcade-Learning-Environment/blob/40c01ba46594f6dc416f4242fe3a76380e66e0fc/.github/workflows/build-ale-py.yml#L47-L50

So now the PyPi test wheels will look like: `ale_py-v0.6.0.devXX+SHA`.

.github/workflows/build-ale-py.yml
mgbellemare: Hi @JesseFarebro , can this be merged?

JesseFarebro: Not yet, still waiting for a release of `cibuildwheel`. I'll make sure to let you know when this can be merged.

justinkterry:
@JesseFarebro @christopherhesse Is this in a state where it can be merged? This is relevant to my work and I don't want it to stagnate.

JesseFarebro:
Still waiting on an official release of `cibuildwheel` with Github action support. Gym supports ALE 0.6, if you need anything newer you can just build it yourself from the repository.

christopherhesse:
:tada:

PR #312, opened 2020-02-21
[games] Stylistic cleanup

tkoeppe:


tkoeppe:
Hang on, there's a lot more of this we can do.

PR #323, opened 2020-03-23
[games] Functional enhancements

tkoeppe:


src/games/supported/LostLuggage.cpp
```diff
@@ -61,24 +61,6 @@ bool LostLuggageSettings::isMinimal(const Action& a) const {
   }
 }
 
-bool LostLuggageSettings::isLegal(const Action& a) const {
```
mgbellemare: I don't think this should be removed -- it will reset the game mid-way through.

tkoeppe: I see. I'm removing this commit from the pull request. We should discuss offline.

PR #327, opened 2020-03-25
[games] Several new games

tkoeppe:
Your careful review would be greatly appreciated.

### DIFF
--- a/src/games/Roms.cpp
+++ b/src/games/Roms.cpp
@@ -22,49 +22,69 @@
 #include "supported/Asterix.hpp"
 #include "supported/Asteroids.hpp"
 #include "supported/Atlantis.hpp"
+#include "supported/Atlantis2.hpp"
+#include "supported/Backgammon.hpp"
 #include "supported/BankHeist.hpp"
+#include "supported/BasicMath.hpp"
 #include "supported/BattleZone.hpp"
 #include "supported/BeamRider.hpp"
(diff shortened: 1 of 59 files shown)


JesseFarebro:
@tkoeppe Can you run Backgammon? It hangs up for me, I haven't investigated much beyond this.

tkoeppe:
@JesseFarebro: I believe we've tried all the game modes that the support code advertises. Backgammon only appears to have one valid difficulty, which is "3"; did you set that?

JesseFarebro:
The code block linked [here](
https://github.com/tkoeppe/Arcade-Learning-Environment/blob/e6776698d22c6e866c385cf70e0a5ac6939521e1/src/games/supported/Backgammon.cpp#L87-L102) ends up in an infinite loop. Specifically, `char address = 0x81;` will be negative because it's signed (`0x81 (129) > 127`), thus `address` can never be `> 0x8d (141)` (which is interpreted as the integer value) so the loop never ends. `address` will continue to loop over `[-127, 127]` never surpassing `141` thus never terminating. Same thing for the second loop. Any idea what might have happened here?

tkoeppe:
Ah -- our chars are unsigned, so this may be a programming error and we should really use `unsigned char` there!

tkoeppe:
#335 should fix this. Thanks for reporting!

PR #328, opened 2020-03-26
CMake modernization

JesseFarebro:
# CMake

The old CMake build was ageing, to say the least. I rewrote the entire build system to conform to more modern CMake paradigms. Here is a quick overview:

## Improvements

* No more global `set`, `add_`, etc. It all uses targets and inheritance now.
* Uses a single `ale` object target so we don't recompile the same files for different upstream targets. This greatly improves compile time.
* Does away with a global CMakeLists.txt and splits things into more modular CMakeLists.txt in each subdirectory.
* Installation is much simplified and will work better with C++ projects using CMake.
* The version string now has one source of truth (CMakeLists.txt), I configure a `version.hpp` through CMake so we no longer have that hard coded string in `ale_interface.hpp`.

## Significant Changes

* I got rid of the old Makefiles and the `module.mk` files. The makefiles didn't work and hadn't been touched in a while.
* I changed all relative imports to absolute imports from the source root, e.g., `#include "../RomSettings.hpp"` -> `#include "games/RomSettings.hpp"`
* I moved the entire `m6502` tree to `emucore`, this is what upstream Stella has done and it makes imports work out better. No more `emucore/m6502/src/bspf/src/...`
* All the licenses for Stella, m6502, bspf, SDL, etc. are now in a common `LICENSES.md` file while the main GPLv2 License is in `LICENSE.md`


mgbellemare:
@JesseFarebro It looks like there are some conflicts now. Would you be able to fix these yourself?

Also, I love the cleaning you've done here. I'm wondering if this wouldn't warrant a new version (0.7). Perhaps once @tkoeppe is also done with his changes? Especially if some of the interfaces are deprecated, and some new functionality is created.

JesseFarebro:
Rebased on master. I changed the relative imports of #327 to absolute imports.

tkoeppe:
@mgbellemare: I have nothing else staged for now, thanks. As to this PR; well, it's your repository and your decision! I'd probably prefer this broken up into many smaller, single-purpose changes, and I feel a little overwhelmed by it in its current form. Any API breaks would of course affect your users, so whether and how you do that is between you and your users, I suppose. Personally, I'm not sure if my use case would be willing to go along with this change (as opposed to just drifting out of sync), but that's based on the sheer size and impact of the change, not on a detailed review. Small, single-purpose changes would be much much easier to appraise.

JesseFarebro:
@tkoeppe I'll break this up into smaller chunks and we can go from there.

tkoeppe:
Thanks! (But feel free to force-push and retain this PR!)

tkoeppe:
(Examples of the kind of trivial changes that would probably easily fly through in isolation: deletion of patently unused build files, build system modernisations that have no visible impact, documentation updates. Then, something like changing all the header inclusions should probably be the _only_ change in its commit -- it has extremely wide impact, so it should be completely mechanical.)

JesseFarebro:
@tkoeppe these changes are now split up into rather verbose commits. If you had the time I'd love to hear your feedback.

tkoeppe:
Thanks! Could you perhaps copy the commits that are obviously trivial (like "Remove old build system: module.mk files, makefiles" and "Remove TinyMT") into a new PR, so hopefully we can land those immediately (@mgbellemare?)? That would further undistract from the substantial changes here.

tkoeppe:
What's the motivation for "Move m6502 -> emucore, bspf -> emucore"? I'm not sure why one would want to "follow Stella upstream" when there are no plans otherwise to actually update Stella. As it is' ALE's stella is some seven (?) years behind upstream, with no plans for an update (which would surely have some behavioural differences). Is it worth churning in-place just for the sake of it?

tkoeppe:
Re "Remove controllers & ale cli": is that also trivial and obvious? Could you add a bit of rationale to the commit message? (Maybe also to the other commit messages, wherever it makes sense?)

JesseFarebro:
@tkoeppe we now have #331 which addresses the removal of files independent of the new build system. #332 which deals with deprecating controllers & updating the docs.

Now the only thing I have in this PR is the new CMake build, I'll leave out the ale-py pybind11 interface for another PR once this is dealt with.

Re moving m6502 & bspf: since I now have CMakeLists.txt in subdirectories managing the inclusion of source files it was a little messy to have `src/emucore/CMakeLists.txt`, `src/emucore/m6502/src/CMakeLists.txt`, `src/emucore/m6502/src/bspf/src/CMakeLists.txt`. Now granted I could have just included those files from `src/emucore/CMakeLists.txt` but I wanted to be consistent and I thought `#include "m6502/src/bspf/src/bspf.hxx` looked ugly anyway. This led me to see what Stella had done upstream and I just decided to adopt that here.

Re plans to update Stella: I actually have PRs for Stella that would make it a little nicer for use cases like the ALE and LIBRETRO. I have a half working version of the ALE with upstream Stella but I haven't discussed anything with @mgbellemare. I'm not sure if this is even something he's interested in. I was wanting to do some benchmark comparisons to see if the performance improvements are worth it.

JesseFarebro:
I rebased this off of #332. @tkoeppe do you have any objections to the state of this PR? It's fairly straight forward now.

RE upstream Stella: I hacked together a minimal version of the ALE based on upstream Stella. One thing I soon realized is they moved to cycle-exact emulation and performance is ~10x worse. The last version they had before this change was 4.7, we use 2.4. Between 2.4 and 4.7 there aren't many things that would benefit us apart from native digital paddle control. Probably not worth the work.

justinkterry:
@tkoeppe How are you feeling about this now?

tkoeppe:
I'm glad we've extracted some small changes, but I don't really have an opinion on this remainder. It's obviously a very large change, but I think this is Marc's call... Personally, I welcome a simplification of include paths (getting them all to start at the same relative root), but I'm not too wild on moving files around just for the sake of it.

justinkterry:
@mgbellemare what do you think?

CMakeLists.txt
```diff
@@ -1,186 +1,11 @@
-cmake_minimum_required (VERSION 2.6)
-project(ale)
-set(ALEVERSION "0.6.0")
+cmake_minimum_required (VERSION 3.14)
 
+project(ale VERSION 0.6.1
```
mgbellemare, line 3: Is this version number picked up by version.hpp.in?

JesseFarebro: Yes, `configure_file ("version.hpp.in" "version.hpp")` in CMakeLists provides the `@PROJECT_VERSION@` replacement.

mgbellemare:
I skimmed the PR but haven't managed to grok it yet. A lot of files seem to be changed at once. It's called "Cmake modernization" but seems to do much more -- I tend to agree with @tkoeppe , is there no way to do this progressively?

JesseFarebro:
Sure, I'll boil this down to **just** the CMake changes without touching relative imports or the directory structure. I'll make new PRs for these after the CMake changes are dealt with.

JesseFarebro:
@mgbellemare this PR now has the minimal CMake changes for the current directory structure and doesn't change any of the relative imports. It should be easier to review now.

src/ale_interface.hpp
```diff
@@ -33,23 +33,25 @@
 
 #include "emucore/FSNode.hxx"
 #include "emucore/OSystem.hxx"
+#if (defined(WIN32) || defined(__MINGW32__))
 #include "os_dependent/SettingsWin32.hxx"
 #include "os_dependent/OSystemWin32.hxx"
+#else
 #include "os_dependent/SettingsUNIX.hxx"
 #include "os_dependent/OSystemUNIX.hxx"
+#endif
```
tkoeppe, line 42: Are these conditionals needed? How come it works at the moment?

tkoeppe:
Thanks, much better! The way the PR looks now I have essentially no opinion on it. If Marc finds it useful, then I'm perfectly happy with it. @mgbellemare: your call!

justinkterry:
@mgbellemare how are you feeling about this now?

weepingwillowben:
I merged this into my fork of ALE, and it generally looks good.  Certainly convenient to cut down compilation times by a factor of 3. 

However, there is an issue with the CMakeLists.txt file that is also a problem with the old version, but now is as good a time to fix it as any. When I run it, the executable crashes on `std::call_once` because the pthreads library is not linked. This is a very annoying and hard to track down bug that can be fixed with 

```
find_package(Threads REQUIRED)
target_link_libraries(ale PRIVATE Threads::Threads)
```

tkoeppe:
@weepingwillowben: Thanks! Yes, GCC (and Clang?) need `-pthread`; if there's a good CMake way to enable this, that would be great.

JesseFarebro:
You can use `FindThreads` which won't return anything if it's provided by the system. I'll update this PR to address this.

weepingwillowben:
So this doesn't seem to be compatible with the current setup.py. In particular, setup.py crashes when looking for the version number in CMakeLists.txt, and then there are probably more issues after that?

JesseFarebro:
You can look here for the Python interface updates: https://github.com/JesseFarebro/Arcade-Learning-Environment/tree/pybind11

JesseFarebro:
The threads issue is now resolved in this PR. Everything should be good, @mgbellemare you can take another look.

mgbellemare:
LGTM!

PR #331, opened 2020-03-30
Remove old/unused files

JesseFarebro:
This PR removes:

1. Old build files, i.e., module.mk files, makefiles, common.rules
2. TinyMT, we now use STL random
3. misc_tools.h, this hasn't been used since v0.5


JesseFarebro:
This can be merged irrespective of #328.

PR #332, opened 2020-03-30
Deprecate controllers, cli & update examples to reflect this

JesseFarebro:
- Removed RLGlue interface. Its own website says it's deprecated, it hasn't been updated in over a decade and I searched Github and couldn't find a project which actually used it.

- Removed FIFOInterface. It seemed to be in a dire state, the examples didn't work, the interface segfault's, and I couldn't find a single project using this interface. This also means I removed the Java example we previously had.

- Since I removed both the interfaces I also removed the CLI build along with its corresponding functions in emucore/Settings.


JesseFarebro:
The checks will fail as this builds off of #328.

JesseFarebro:
This can be merged, but for the builds to succeed we would need #328 first.

tkoeppe:
The it can't really "be merged" :-) Which specific part of #328 does this depend on?

JesseFarebro:
I misspoke. This doesn’t intrinsically depend on #328. I could change the current CMakeLists.txt to no longer build the new examples & cli but this would be unnecessary due to #328. 

I should have submitted this after #328 is dealt with. I wish Github could specify dependent PRs.

tkoeppe:
Ah. Alternatively, you could rework this PR to work independently, and instead add the relevant opposite changes to #328? Assuming that the rest of this PR is obvious and trivial?

JesseFarebro:
Sure, I’ll go ahead and do that.

tkoeppe:
Thanks! (I welcome these deletions, they definitely make it easier to keep our local checkout in sync!)

JesseFarebro:
@tkoeppe this now has the minimal needed changes to `CMakeLists.txt` so we can merge this independently of #328.

tkoeppe:
Sounds good, thanks! @mgbellemare: what do you think?

mgbellemare:
It feels wrong to kill the command-line altogether, but I can be convinced. Does that mean that running the ALE requires either

- Linking to ale_interface (C++)
- Linking ale_interface.py (Python)

? And as a consequence, no more support for other languages?

I might recommend keeping the CLI and 'Usage' printout as a stub, at the very least saying this ("The command-line interface is deprecated. To use ALE, you must do one of two things: ...").

JesseFarebro:
Yes,

1) Linking to the C++ library
2) Using ale-py, or building the Python library yourself.

No other language support right now. I've searched over Google, Github, etc. and I can't find any recent usage of the fifo or RLGlue interface that is **public**. Do you have any idea of the demand for other languages?

Would you be opposed to removing the cli but keeping `-DBUILD_CLI` in cmake and printing the deprecation message if the flag is enabled? I guess I could do that for `-DBUILD_RLGLUE` as well?

mgbellemare:
I think if you don't build the CLI, might as well remove it altogether. My thinking is that there should be a clear entry point for new users. Agreed in this day and age a library is more likely. Will you/have you update/d the README.md to reflect this?

JesseFarebro:
@mgbellemare so you are okay with this PR how it is? Getting rid of the CLI entirely? I'm not sure what you mean by a clear point of entry for new users.

I have prepared some documentation but it also includes some changes in #328 so once we have all these PRs sorted I will update the docs.

mgbellemare:
Yes, go ahead.

tkoeppe:
@JesseFarebro: By the way, do you care about merging and supporting divergent forks? Otherwise, would you consider just using squash or rebase to apply pull requests, so as to line up the commits in order and not have the extraneous merge commits around?

JesseFarebro:
Sure, I’ll do that from now on.

tkoeppe:
Great, thanks!

Issue #336, opened 2020-04-02
Missing checksums for new game ROMs


justinkterry:
@tkoeppe The 30 new games you added didn't get MD5s for them added to the list: https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/md5.txt

This is needed to verify you're using the right ROM for the code. Can you please look into this?

justinkterry:
It turns out all the relevant md5s checksums are already in here, but elsewhere?:

https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/emucore/stella.pro

tkoeppe:
I'll see that we collect those checksums; but the verification is not fatal, right?

justinkterry:
It's not fatal. Someone who works with me also submitted a PR for this- #338

PR #338, opened 2020-04-03
updated md5 file

MarioJayakumar:
The md5.txt file has been updated to include md5 hashes of the new ROMS supported by ale.


justinkterry:
This closes issue #336

justinkterry:
@tkoeppe can this be merged? It's a pretty straightforward fix.

JesseFarebro:
I think more thought has to go into this and what hashes we standardize on. For example, Wizard of Wor in this MD5 list is the NTSC version, the current version used by Gym is the PAL version. Resized frames will have different proportions and colours.
In hindsight, this should have probably been standardized earlier. @mgbellemare if we are going to force ROM hashes should we standardize on what Gym has right now? That way results will be comparable with Dopamine, etc. I would imagine most research was conducted with Gym.

RE this PR: @tkoeppe do you have any updates on what hashes were used internally?

tkoeppe:
@JesseFarebro: yes, sorry, here it is: https://github.com/mgbellemare/Arcade-Learning-Environment/pull/343

(I don't have a particular opinion on whether we agree on any particular ROM; as I said, we ignore the checksums anyway.)

justinkterry:
I think we really need to standardize the ROMs. In virtually all retro games, there almost certainly are small regionally variations, and new ones are regularly being found. This is sure to cause trouble.

Additionally, not being able to do checksums to make sure a ROM is suitable, while expecting end users to drop in their own ROMs, is also asking for trouble.

justinkterry:
My vote would be to specifically to use Gyms ROMs for games in Gym, and to use thomas' ROMs for the newly added games. I believe there are ROMS without MD5s that aren't in Gym and were added before Thomas' addition though. @JesseFarebro any idea what the best course of action on those would be?

JesseFarebro:
I wouldn't think imposing a hard restriction in the ALE is the way to go. We should obviously fix the warning message. Perhaps we could be louder about the warning, but at the end of the day, I think a hard restriction on ROMs should be implemented in Gym.
This is already what happens, and when ale-py replaces atari-py the end-user ROM import process will be implemented in Gym thus makes it easier to catch invalid ROMs there.

justinkterry:
@JesseFarebro That makes sense. We'd still need a "correct" list of hashes to put into ALE-Py, even if it's just a warning, right?

JesseFarebro:
Yes, I'll close this in favour of @tkoeppe's PR and then I'll create a PR resolving the Gym MD5 vs our MD5.
There will be additional work after that to make sure we are properly packaging and use the MD5 list.

> I believe there are ROMS without MD5s that aren't in Gym and were added before Thomas' addition though.
@justinkterry what do you mean by this? Can you elaborate?

PR #344, opened 2020-04-20
Remove redundant namespace qualifications and stream flushes

tkoeppe:


tkoeppe:
The `ale::` qualifications were left over from before we moved everything inside namespace `ale`.

Issue #346, opened 2020-04-28
wizard of wor not working


weepingwillowben:
I am trying to run the games and make sure they are controllable, and Wizard of Wor is not for some reason. 

I am choosing random actions and the behavior I actually see is as if I am continually choosing the no-op action. The player eventually gets kicked out of the waiting area and eventually gets killed by the NPCs without ever firing a single bullet or moving anywhere. 

Is there some trick to getting this game working? I have gotten over a dozen other games working just fine. 

Here is the code I am running:

```python
import ale_py
from random import randrange
import pygame
import numpy as np
import time

pygame.init()

ale = ale_py.ALEInterface()

BOX_PIX = 3
BSX = 160
BSY = 210
screen = pygame.display.set_mode((BSX*BOX_PIX, BSY*BOX_PIX))

ale.loadROM("wizard_of_wor.bin")

legal_actions = ale.getMinimalActionSet()

for episode in range(10):
  total_reward = 0
  while not ale.game_over():
    a = legal_actions[randrange(len(legal_actions))]

    for _ in range(5):
        reward = ale.act(a)
        total_reward += reward
    time.sleep(0.02)
    image = ale.getScreenRGB()
    image = np.transpose(image,(1,0,2))
    #print(image.shape)

    myImage = pygame.image.fromstring(image.tobytes(), image.shape[:2][::-1], "RGB")

    #surf = pygame.surfarray.make_surface(image)
    screen.fill((0,0,0))
    screen.blit(myImage, (0, 0))

    pygame.display.flip()

  ale.reset_game()
print(total_reward)
```

JesseFarebro:
I've noticed this myself as well. I believe this issue is not just related to Wizard of Wor. I will take a deeper look at this soon and get back to you.

JesseFarebro:
PR #347 solves this issue. There hasn't been a release since this issue was introduced so the impact is minimal.

weepingwillowben:
Thanks for looking into this. Indeed, this seems to fix the issue.

weepingwillowben:
Closing because this issue has been fixed.

PR #347, opened 2020-04-28
SwapPort Fix

JesseFarebro:
@tkoeppe is there any reason that the default value of `sp` was `NO`? The sp check in `OSystem` for swap port looks like:

https://github.com/mgbellemare/Arcade-Learning-Environment/blob/186abae7f10826f4b223f825b560ad05fc7dd32d/src/emucore/OSystem.cxx#L582-L583

This would mean that the ports were always being swapped and this impacted games such as Wizard of Wor in #346.


tkoeppe:
@kieranhj: I think this is your code, could you perhaps review this?

JesseFarebro:
@kieranhj any update on this?

kieranhj:
Apologies, the alert for this mention went into the spam folder of my personal email.

The default to "NO" was because of this line: https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/emucore/Console.cxx#L85

But it looks like I misinterpreted its use? A setting has to be explicitly added to the list for it to be configurable later on by the ROM Settings, so needs a default value. I'll need to go back and check whether Surround still works with the new default of "".

justinkterry:
@kieranhj did you get a chance to check out surround?

JesseFarebro:
The problem is as follows:

Games such as Wizar of Wor need to swap their ports. This is done via `stella.pro` and then we generate `DefProps.hxx`.
By changing the default value of the swap port we override the ROM specified value of swap port defined in `stella.pro`, hence why any game that requires a swap port wasn't working.

It seems Surround is the only game that performs a swap port outside of stella.pro. I think the proper way to solve this is to manually set `"Console.SwapPorts" "YES"` on the two known Surround ROMs in our stella.pro file. We would then need to regenerate `DefProps.hxx` and it should work as expected without messing with the default value of swap port.

JesseFarebro:
This PR now removes manual swap port modification performed by Surround. This, in turn, allows us to remove the default swap port value in `Settings.hxx`.

I also modified `stella.pro` to enable SwapPort for all known Surround ROMs in our database. I then regenerated `DefProps.hxx` so these changes are picked up.

@tkoeppe @kieranhj could you take a look now?

I verified that Surround and other SwapPort ROMs do indeed perform a SwapPort as expected.

kieranhj:
Thanks Jesse. This seems like a much more sensible and elegant solution to the SwapPorts problem for Surround than overriding in code.

tkoeppe:
Thanks everyone, that's great! @mgbellemare: This seems to be a big improvement and we'd be hapy to have it committed.

mgbellemare:
Sg, thanks @JesseFarebro !

Issue #345, opened 2020-04-27
You accidentally put roadrunner in md5.txt twice


justinkterry:
@tkoeppe in PR #343 you accidentally added two different MD5s for roadrunner:

```
2bd00beefdb424fa39931a75e890695d road_runner.bin
ce5cc62608be2cd3ed8abd844efb8919 road_runner.bin
```

JesseFarebro:
Only recently Gym patched the NTSC version (`ce5cc62608be2cd3ed8abd844efb8919`) to match the PAL version (`2bd00beefdb424fa39931a75e890695d`). I don't even think the PAL version has been deployed in Gym as they still use an older version of atari-py.

It's probably safe to go with the NTSC version `ce5cc62608be2cd3ed8abd844efb8919`, the one @tkoeppe added. It seems difficult to find the PAL version.

justinkterry:
Do you want me to submit a small PR to remove the PAL version then?

hholst80:
Certainly they should not have the same filename. Do not reinvent the wheel. Follow the TOSEC naming schema if you are going down this path with multiple versions of the same game.

https://www.tosecdev.org/tosec-naming-convention

JesseFarebro:
@hholst80 the MD5 list doesn't support multiple versions of the same game, this was simply a mistake. The MD5 list is so we can at least attempt to have a single source of truth for supported ROMs in the ALE so researchers can compare results.

@justinkterry I'm not even sure anymore, upon further investigation Gym uses a different version entirely:
```
MD5 (road_runner.bin) = 7d3cdde63b16fa637c4484e716839c94
```

The ROM info is:
```
"Cart.MD5" "7d3cdde63b16fa637c4484e716839c94"
"Cart.Manufacturer" "CCE"
"Cart.Name" "Road Runner (CCE)"
```

From what I can tell CCE produced games in Brazil mainly. I still think going with the official Atari NTSC version `ce5cc62608be2cd3ed8abd844efb8919` is probably the right way to go.

tkoeppe:
Sorry about this. I've asked for our situation to be clarified, but as always, we have no deep interest in this checking mechanism, and if there's a particular thing you prefer, feel free to do that.

justinkterry:
@JesseFarebro Gym using the brazilian version seems like a problem on Gym's end..?

JesseFarebro:
[...]

PR #348, opened 2020-05-01
Remove duplicate roadrunner md5

justinkterry:
Closes issue #345 in the manner agreed upon in the discussion.


JesseFarebro:
As per the discussion in #345, our current roadrunner hash `2bd00beefdb424fa39931a75e890695d` is hard to find and support got dropped from upstream Stella? OpenAI uses hash `7d3cdde63b16fa637c4484e716839c94` which is the NTSC version published by CCE. We will standardize on hash `ce5cc62608be2cd3ed8abd844efb8919` which is the official Atari NTSC version of road runner.

PR #353, opened 2020-05-15
Source file path resolution

JesseFarebro:
The justification for these changes is for ease of maintainability and making it easier for contributions to the ALE. Before, it was pretty hard to tell what's going on with imports.

## Moving m6502 and bspf
It seems wrong to have the directory structure: `src/emucore/m6502/src/bspf/src`. The `src` directories are redundant and it's just unnecessary. Stella now uses the `emucore` to host the M6502 files. I just adopted their changes. So to recap the changes are:

- `src/emucore/m6502/src/**` -> `src/emucore`
- `src/emucore/m6502/src/bspf/src/**` -> `src/emucore/bspf`

## Resolve relative imports

I resolved all imports of the form `#include "../*"` to be relative to the src directory. So for example, `#include "../RomUtils.hpp"` -> `#include "games/RomUtils.hpp"`.

## All imports being relative to src

I resolved all imports to be relative to src instead of having a search path in every submodule. For example, there used to be imports in `emucore` for files in `common` without specifying the submodule. For example, `#include "SoundSDL.hxx"` now becomes `#include "common/SoundSDL.hxx"`. Before you'd have no idea where `SoundSDL.hxx` came from at first glance.

This makes it much easier to grok where things are being resolved. The old way of using these global include paths didn't make much sense to me.

Almost every file is impacted but the changes are minimal. I think this is a much-needed quality of life improvement. Working with the codebase was a struggle at times with the current module resolution.

@tkoeppe I'd love to hear your feedback as you seemed reluctant to these changes in the other PR. I think things are more modular now and hopefully, it doesn't look too daunting at first glance.


JesseFarebro:
@tkoeppe any thoughts on this now?

tkoeppe:
Ah, hello, I had missed the last ping! Is the description still accurate, i.e. are you still moving any files around? I do welcome the change of search paths that makes all inclusions relative to a single, common root, that's a big win.

JesseFarebro:
The description is still accurate. The only files that have moved are:
`src/emucore/m6502/src/*` -> `src/emucore`
`src/emucore/m6502/src/bspf/src/*` -> `src/emucore/bspf`

As we had discussed before my main motivation for this was to get rid of the nested `src` directories as I thought it was redundant and unnecessary. Hopefully, you don't have too many internal files that would need a path change. If so, this commit was generated with a script I wrote, I could send it your way if you wanted.

tkoeppe:
OK. And you really want that file move? I suppose that's fine, though can we make that in a separate commit?

JesseFarebro:
It's not entirely necessary but I just thought the readability improvements alone, e.g., `emucore/bspf/bspf.hxx` instead of `emucore/m6502/src/bspf/src/bspf.hxx` would be worth it in and of itself.
If you're vehemently opposed to moving these files I would be fine to revert this change.

Regardless, I will make it a separate commit whichever way we go. Thanks for taking a look!

tkoeppe:
I'm not violently opposed, I just mildly feel that this is unnecessary churn. Header inclusions aren't big readability pain points in my view, so the improvement is rather slight. I have no problem adopting this change if you want it, I just feel like I'd rather be doing anything else (e.g. specing out the multi-controller API). By contrast, having the single search path root is a real advantage in terms of _maintenance_, since it simplifies the build system and actually improves readability (since I don't need to mentally resolve `..` etc., which requires knowing the path of the current file; that's a very real cost).

There's perhaps an argument to be made for (partially?) tracking the upstream Stella changes, though I really have no idea how impactful that is -- it could go all the way from positive (because it helps identify the current version of the file) to neutral (we'll replace everything anyway) to negative (my three-way merge is now broken), and there's a very real chance we'll never ever upgrade Stella, so... I would probably prefer to just not have to think about this at all. But that's just my personal view. If you really like this change, don't let me stop you.

JesseFarebro:
@tkoeppe so all the m6502 file changes are in a separate commit that could be cherry picked from this set. I'd still like to include it though.

It's also worth noting here that we will never track upstream Stella. They moved to cycle-exact emulation shortly after the ALE diverged. The side effect of cycle-exact is that emulation speeds are significantly slower, around 10x slower in my benchmarks.

tkoeppe:
OK, thanks -- so the question of what Stella code is doing seems to be irrelevant then. I have nothing to add then, the first paragraph of my last comment still applies: I find the uniformization of the include path very valuable, and the moving around of files churn. But that shouldn't stop you, it's your repo and if you find this an improvement, it's your call. I can happily live with the changes. If you could layer the commits so tha the code move comes last (and so that I can omit that), I'd welcome that.

Thanks!

JesseFarebro:
I moved the file changes to the last commit. Thanks for the feedback on this PR!

Issue #337, opened 2020-04-03
Md5 warning emitted all the time in ale_py


weepingwillowben:
So when the ale_interface is used to load a ROM, it checks for a `md5.txt` file in the current working directory here:
https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/ale_interface.cpp#L76

before emitting a warning here:

https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/ale_interface.cpp#L112

This is not optimal, as it spits out warning whenever that md5.txt isn't around, which is typically 100% of the time if you are using ale_py. 
@JesseFarebro

justinkterry:
To fix this, the md5.txt would have to be included with the PyPI release, and the code would have to be refactored to check in that directory. Having the check be for the current working directory seems odd to me.

JesseFarebro:
Yeah, I agree. It's been like this for a long time, I think a better solution is to actually do md5 validation when loading the ROM. Right now we just check against a filename, e.g. `Freeway.bin` won't load but `freeway.bin` will.

@mgbellemare was there a reason you didn't want to enforce md5 validation?

tkoeppe:
(Same agreement here; the overly simplistic file handling logic made it impractical for us to try and meaningfully distribute and use the checksums. We just ignore the warning. A proper solution would perhaps a) make the path configurable via flag or setting, or b) hardcode the numbers in the program code.)

justinkterry:
Is the plan to handle this logic here, or in Gym?

JesseFarebro:
In the ALE I'm thinking, once #328 is dealt with I will make sure the md5 list is packaged properly so we don't have these extraneous warnings.

JesseFarebro:
All warnings regarding MD5 mismatches can now be taken seriously with #356. There are no breaking changes, and loading ROMs is now more robust. Both the old snake case naming convention as well as MD5 matching will be valid from now on.

Issue #349, opened 2020-05-01
Verified MD5 hashes of several games


MarioJayakumar:
For Asteroid, Blackjack, Lost Luggage, Private Eye, and Space War, I am currently unable to find ROMs that have the same MD5 hash as those specified in md5.txt. I have tried several different ROMs on various sites, using both NTSC and PAL, and yet none of the ROMs have a correct hash. I also saw in issue #345  that these ROMs may potentially be using a completely different version like Road Runner was. Could you verify that the MD5 hash that is listed is indeed correct and if so what format, language, or version that correct ROM is using? I am working on this with Justin Terry and Ben Black.

justinkterry:
@JesseFarebro

JesseFarebro:
Regarding Asteroids and Private Eye, OpenAI uses the same hash as we do so I'd be hesitant to change this. Regardless, all of these ROMs can be found, for example, you can check out the ROM download instructions on https://github.com/openai/atari-py.

Hope this helps.

justinkterry:
As far as we can tell, ROMs matching ALE's MD5s are only available in that rar file. We can't find them anywhere else, even on atarimania. Surely this is undesirable?

MarioJayakumar:
To add on to this, I compared the MD5 checksum of the ROMs included in the RAR file and all of  hashes match what we want. For example the correct version of Asteroids is available in this RAR file, as a ROM from Sears. However on Atarimania's page for Asteroids, the Asteroids ROM with this same serial and also from Sears does not have a matching MD5 hash. This indicates that either the link on Atarimania is broken or that the version distributed in the RAR is not available elsewhere on the site.

JesseFarebro:
Are these 5 ROMs the only you can't find through other sources?

I'd have to take a deeper look, could you do a diff to get a sense of how much they differ?

MarioJayakumar:
Yes these 5 ROMs are the only ones I could not find on any other website, apart from this RAR. Are you saying that I should do a diff of each version of each ROM?

JesseFarebro:
What I meant is maybe start by doing a diff of the two Asteroids ROMs from Sears and see how much they differ. I'm curious about what changes between two seemingly equivalent ROMs by the same distributor. I wonder where Stella has compiled the `stella.pro` file from, perhaps there is more information to be had.

MarioJayakumar:
[...]

PR #356, opened 2020-05-24
Added functional md5 checking

benblack769:
Hardcodes all md5 values into ALE. addresses #337 

Now ALE searches for an ALE game setting with a md5 that matches the ROM setting 
 when roms are loaded.

For backwards compatibility, after the md5 check fails, the old filename
check initiates, and if a matching filename is found, the old warning is generated, along with
an additional message saying what the md5 should be.

Also adds an isSupportedRom method to the interface so that the user can 
verify forcibly whether the rom md5 matches the expected md5, rather than just getting a warning.

### DIFF
--- a/src/ale_interface.cpp
+++ b/src/ale_interface.cpp
@@ -41,11 +41,13 @@
 #include <sstream>
 #include <stdexcept>
 #include <vector>
+#include <fstream>
 
 #include "common/ColourPalette.hpp"
 #include "common/Constants.h"
 #include "emucore/Console.hxx"
 #include "emucore/Props.hxx"
+#include "emucore/MD5.hxx"
 #include "environment/ale_screen.hpp"
 #include "games/RomSettings.hpp"
 
@@ -73,21 +75,6 @@ void ALEInterface::createOSystem(std::unique_ptr<OSystem>& theOSystem,
   theOSystem->settings().loadConfig();
 }
 
-bool ALEInterface::isSupportedRom(std::unique_ptr<OSystem>& theOSystem) {
-  const Properties properties = theOSystem->console().properties();
-  const std::string md5 = properties.get(Cartridge_MD5);
-  bool found = false;
-  std::ifstream ss("md5.txt");
-  std::string item;
-  while (!found && std::getline(ss, item)) {
-    if (!item.compare(0, md5.size(), md5)) {
-      return true;
-    }
-  }
-
-  return false;
-}
-
 void ALEInterface::loadSettings(const std::string& romfile,
                                 std::unique_ptr<OSystem>& theOSystem) {
   // Load the configuration from a config file (passed on the command
@@ -109,19 +96,6 @@ void ALEInterface::loadSettings(const std::string& romfile,
     Logger::Error << "ROM file " << romfile << " not found." << std::endl;
     std::exit(1);
   } else if (theOSystem->createConsole(romfile)) {
-    if (!isSupportedRom(theOSystem)) {
-      const Properties properties = theOSystem->console().properties();
-      const std::string md5 = properties.get(Cartridge_MD5);
-      const std::string name = properties.get(Cartridge_Name);
-
-      // If the md5 doesn't match our master list, warn the user.
-      Logger::Warning << std::endl;
-      Logger::Warning << "WARNING: Possibly unsupported ROM: mismatched MD5."
-                      << std::endl;
-      Logger::Warning << "Cartridge_MD5: " << md5 << std::endl;
-      Logger::Warning << "Cartridge_name: " << name << std::endl;
-      Logger::Warning << std::endl;
-    }
     Logger::Info << "Running ROM file..." << std::endl;
     theOSystem->settings().setString("rom_file", romfile);
   } else {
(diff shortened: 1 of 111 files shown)


src/games/Roms.cpp
```diff
@@ -229,14 +233,23 @@ static const RomSettings* roms[] = {
     new ZaxxonSettings(),
 };
 
-/* looks for the RL wrapper corresponding to a particular rom title */
+/* looks for the RL wrapper corresponding to a particular rom md5,
+then title in that order. returns null if neither match */
 RomSettings* buildRomRLWrapper(const std::string& rom) {
+  std::ifstream romfile(rom);
```
JesseFarebro, line 239: This file loading doesn't seem necessary. I would refactor `ALEInterface::loadSettings` to remove the ROM validation check. You should then be able to perform `ALEInterface::loadSettings` in `ALEInterface::loadROM` before calling `buildRomRLWrapper`. Once we've loaded the settings we'll already know the MD5 of the ROM so we don't need to do this extra work. Hopefully that makes sense.

src/ale_interface.cpp
```diff
@@ -217,6 +218,17 @@ void ALEInterface::loadROM(std::string rom_file) {
 #endif
 }
 
+bool ALEInterface::isSupportedRom(){
```
JesseFarebro, line 221: I don't understand the point of this, you call `isSupportedROM` after you've already loaded a ROM and presumably been warned if it was or wasn't supported? If this is going to be a public API shouldn't it take a filename and actually check the MD5 against the filename. I'm assuming you want an API like this for your ROM loading package?

JesseFarebro (changes requested):
Thanks for taking the time to contribute this. I didn't check over all the MD5 hashes, have you changed any of them with respect to our current MD5 list, e.g., changes from #349?

src/games/Roms.cpp
weepingwillowben: Makes sense.

src/ale_interface.cpp
weepingwillowben: This makes a lot more sense. I'll make that change.

weepingwillowben:
Yes, this is using the hashes we found in md5.txt in the master branch of the repo. I scripted the process of hardcoding these into the files and tested it by running all the games and making sure the md5s matched, so they should all be correct.

src/ale_interface.cpp
weepingwillowben: Made this a static method so it can be called without loading a ROM

src/games/Roms.cpp
weepingwillowben: So I looked at this further, and I don't think it makes sense anymore. It is a little confusing, but I belive `load_settings` must be called after `romSettings->modifyEnvironmentSettings` which in turn must be called after `buildRomRLWrapper`, making this particular refactor impractical.

I also am not sure of the utility of this change. I mean, loading a ROM is already an extremely expensive operation, I am measuring something like 120ms!!! to load a ROM from the python interface with the following script. I have noticed similar performance from the c++ interace. Surely an additional file load is meaningless in the face of this magnitude of inefficiency.
```
from ale_py import ALEInterface
ale = ALEInterface()
for i in range(100):
    ale.loadROM("wizard_of_wor.bin")
```

weepingwillowben:
@JesseFarebro Can you look at this again?

JesseFarebro:
@weepingwillowben I'll get around to this in the next coupled days. Thanks for your patience.

justinkterry:
@JesseFarebro This PR has now been open for a month. What's up?

justinkterry:
@JesseFarebro Any news here? This PR has been open for 3 months.

JesseFarebro:
I want to deal with #354 first as this PR has some changes to the old Python interface. I'll get around to this after the new Python interface.

weepingwillowben:
I fixed the merge conflicts, added the isRomSupported function to the pybind11 API, and added a unit test to make sure it works.

weepingwillowben:
@JesseFarebro The CI tests are failing on the install step, and I don't understand why.

JesseFarebro:
Thanks @weepingwillowben, as for the CI tests, I've seen those failures before. Microsoft hosts their own package repositories and they seem to have intermittent issues. I'll try restarting the tests tomorrow and if it's still not fixed I'll investigate further.

justinkterry:
@JesseFarebro just following up

justinkterry:
@JesseFarebro All tests are passing now. Can you merge this if you have no other issues with it?

JesseFarebro:
I seen they are now passing, I'll give it a final look later today.

JesseFarebro:
@weepingwillowben can you enable ```Allow edits and access to secrets by maintainers``` so I can push to this branch. I altered this PR a little bit, you guys can take a look to make sure it conforms to your use case.

weepingwillowben:
@justinkterry This option doesn't seem to display for me. Maybe I don't have the right permission. Can you do this?

justinkterry:
I added @JesseFarebro to the repo, I think that should suffice?

JesseFarebro:
That works, but you don't see something on the right side that looks like this?

Screen Shot 2020-10-23 at 5 46 36 PM

justinkterry:
I don't see that

JesseFarebro:
You can take a look at my latest commit. I allow you to pass an md5 when building the RL wrapper but I don't have distinct methods for matching md5 vs. filename. The `isSupportedRom` function can explicitly check whether it was due to a hash match. We don't really care when loading the ROM otherwise.

weepingwillowben:
I am happy with these changes.

JesseFarebro:
So there's one last thing that needs to be mentioned. @weepingwillowben we had discussed moving `loadSettings` to the top of `loadROM` but you had mentioned you thought that wasn't possible due to `modifyEnvironmentSettings`. That's not actually the case (anymore), `modifyEnvironmentSettings` makes settings changes, but those will only be parsed in a call to `StellaEnvironment`. The Stella environment only gets constructed after we've loaded all settings and modified the environment settings. Environment settings should have no effect on creating the console which is what happens in `loadSettings`.

Now, this wasn't always the case. Surround previously modified the swap port setting which did need to be performed before console creation. There were some issues with that approach and I removed that modification in #347. There are no longer any ROMs that modify the environment settings beyond the scope of `StellaEnvironment`. I would argue any further modifications to ROM settings outside of `StellaEnvironment` should be done in `DefProps` as in #347.

JesseFarebro (approved):

JesseFarebro:
Thanks for the work on this @weepingwillowben @justinkterry.

PR #357, opened 2020-05-29
setup.py fix to parse CMakeLists.txt

larsr:
CMakeLists.txt does not define ALEVERSION anymore.  Instead the variable is called 'ale VERSION'.  This patch makes `pip install .` work again by modifying setup.py to parse CMakeLists.txt correctly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,7 +105,7 @@ def _parse_version(filename):
     """
     # Parse version from file
     contents = _read(filename)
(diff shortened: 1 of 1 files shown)


JesseFarebro:
Thanks for taking the time to look into the version parsing issue in `setup.py`. As of right now even with this fix, the Python interface won't build properly. The changes in #354 are necessary to build a functioning Python interface with the new build system. #354 also includes the fix you've outlined.
We should be getting around to this soon, but until then you can build off of that branch or build from the last tagged version `v0.6.1`.

PR #362, opened 2020-08-08
FIX: python pip install

Islandman93:
Adds back support for generating a C shared object file using -DBUILD_C_LIB cmake flag.
Fixes the setup.py to support the new CMake version tag and adds building with multiple cores.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext
+from multiprocessing import cpu_count
 import subprocess
 import os
 import sys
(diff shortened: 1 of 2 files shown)


Islandman93:
Oops, should've looked at open merge requests first. This is unnecessary after https://github.com/mgbellemare/Arcade-Learning-Environment/pull/354 gets merged.

setup.py
```diff
@@ -65,11 +65,12 @@ def build_extensions(self):
             if not os.path.exists(self.build_temp):
                 os.makedirs(self.build_temp)
 
+            print(["cmake", ext.sourcedir] + cmake_config_args)
```
Islandman93, line 68: Erroneous print left in

JesseFarebro:
Hi,

Right now you can install a dev build from here: https://pypi.org/project/ale-py, i.e., `pip install ale-py`, or you can build from the Pybind11 branch (git clone https://github.com/JesseFarebro/Arcade-Learning-Environment -b pybind11) which fixes these issues. We've been reworking how the Python interface works (#354), sorry for the inconvenience.

Issue #114, opened 2015-10-08
C interface on OSX doesn't seem to work with SDL


UserAB1236872:
I'm having trouble getting SDL to run via libale_c.so. I've put reference C code and the stack trace in a gist here:

https://gist.github.com/Jragonmiris/b99b5033c52e1ff6d677

This doesn't happen with sharedLibraryInterfaceExample so it's working properly from C++ for some reason. But I can't get the display window to open via any FFI through the C wrapper, whether it's C, Rust, or Python.

USE_SDL is on (obviously, since the C++ example works properly).

I tried explicitly linking with SDL too, with -lsdl and -lsdl2 (just in case), but it didn't seem to help.

mhauskn:
This is a known issue. In the [python example](https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/doc/examples/python_example.py), which uses the ale-c-lib, it is necessary to use pygame init to proxy-call SDL_main on OSX. I believe I was getting the same NSConsistency error you describe in the gist. I wasn't able to find a way to fix this, hence the pygame workaround.

JesseFarebro:
This should no longer be an issue with the SDL2 upgrade.

PR #368, opened 2020-09-07
SDL2 Upgrade

JesseFarebro:
This PR improves upon the previous screen display in the ALE. This seemed necessary as there should be a native cross-platform way to display the screen especially as most people rely on the Gym interactive visualization and its current implementation is a little hacky. This improves upon the previous method in the following

- SDL primitives rewritten with SDL2
- SDL2 can be loaded dynamically at runtime or linked at build-time. This provides us with the advantage of shipping packages like `ale-py` with SDL. Note: SDL is only dynamically loaded and memory-mapped to the process if `display_screen` is enabled. As far as running the ALE headless there are no changes.
- Basic support for window scaling, this just uses a basic heuristic to find an integer scaling factor that doesn't exceed 30% of the max width or height of the current display. The previous display screen functionality only supported one resolution and this could be quite small on HIDPI displays.
- Proper frame rate handling for NTSC and PAL games.
- Display opens properly on macOS which has been an existing bug for a while (#114). The temporary fix was to import pygame before the ALE but that's no longer necessary.

### Manual mode

As of right now, I removed manual mode. I did this primarily to render at the desired framerate (avoiding frame skips). For people wanting to test out games, they can easily download the latest version of Stella. I believe the main purpose of manual mode was to decide on some of the fixed controller parameters.
As far as recording human trajectories I think this could be easily implemented on top of the ALE and doesn't need to be baked in.

## Limitations

- As part of a larger rewrite of the ALE to enable better serialization this PR doesn't include any type of serialization for the screen. That is if you're displaying the screen and load to a different state the screen won't update accordingly. You'd have to run forward one step to update the screen.

- There's no type of on-screen display to give users an idea of what they can control. Currently, you can increase/decrease the speed with the left/right arrow keys and increase/decrease the sound using the up / down arrow keys. This is printed to the console but it would be nice to have a better UI.

- SDL depends on polling for Window manager events in order for the screen to remain responsive (OS limitation, not SDL). This means if you run a Python shell and don't call `act()` fast enough the SDL window will become unresponsive. There are some possible solutions such as running either SDL or the ALE in another thread so we can always poll for events but this hasn't been explored yet.


src/emucore/Console.hxx
```diff
@@ -136,7 +136,6 @@ class Console
     */
     std::string getFormat() const { return myDisplayFormat; }
 
-  private:
```
JesseFarebro, line 139: Not a mistake, this makes `getFrameRate()` public. This is required for `ScreenSDL`.

.github/workflows/build-ale-py.yml
```diff
@@ -34,17 +34,19 @@ jobs:
...
-        CIBW_MANYLINUX_X86_64_IMAGE: "ghcr.io/mgbellemare/manylinux2014-vcpkg"
-        CIBW_SKIP: "*i686* *win32*"
+        CIBW_MANYLINUX_X86_64_IMAGE: "ghcr.io/jessefarebro/manylinux2014-vcpkg"
+        CIBW_SKIP: "*i686* *win32* pp*"
+        CIBW_ARCHS_MACOS: "x86_64"
         CIBW_TEST_REQUIRES: "pytest"
         CIBW_TEST_COMMAND: "pytest {project}"
-        CIBW_ENVIRONMENT: "ALE_BUILD_VERSION=${{ env.ALE_BUILD_VERSION }} VCPKG_DEFAULT_TRIPLET=${{ matrix.config.triplet }}"
-        CIBW_BEFORE_ALL: "vcpkg install ${{ join(matrix.config.deps, ' ') }} --triplet ${{ matrix.config.triplet }}"
+        CIBW_ENVIRONMENT: "ALE_BUILD_VERSION=${{ env.ALE_BUILD_VERSION }} MACOSX_DEPLOYMENT_TARGET='10.15' VCPKG_DEFAULT_TRIPLET=${{ matrix.config.triplet }}"
```
JesseFarebro, line 42: We have to set the min deployment target on macOS to 10.15 in order to get proper C++17 filesystem support.

.github/workflows/build-ale-py.yml
```diff
@@ -34,17 +34,19 @@ jobs:
...
+        CIBW_BEFORE_ALL: "vcpkg install ${{ join(matrix.config.deps, ' ') }} --triplet ${{ matrix.config.triplet }} --overlay-triplets cmake/custom-triplets --feature-flags=-compilertracking"
```
JesseFarebro, line 43: Without disabling compiler tracking we aren't able to compile anything within manylinux docker containers.

.github/workflows/build-ale-py.yml
```diff
@@ -34,17 +34,19 @@ jobs:
     - name: Build wheels
       uses: pypa/cibuildwheel@v2.0.0
       env:
-        CIBW_MANYLINUX_X86_64_IMAGE: "ghcr.io/mgbellemare/manylinux2014-vcpkg"
-        CIBW_SKIP: "*i686* *win32*"
+        CIBW_MANYLINUX_X86_64_IMAGE: "ghcr.io/jessefarebro/manylinux2014-vcpkg"
```
JesseFarebro, line 37: This must be changed once we have a better solution for managing docker images.

PR #375, opened 2020-11-18
CMake refinement

JesseFarebro:
This PR includes:

- subproject compiler options instead of target based options. This will stop the propagation of compiler options through `_INTERFACE`
- updates to the test framework to use CTest in preparation of C++ tests.
- CI Python bump to 3.9, installing cmake via pip for multiplatform consistency, and pybind11 bump to 2.6.1

### DIFF
--- a/.github/workflows/build-ale.yml
+++ b/.github/workflows/build-ale.yml
@@ -24,10 +24,10 @@ jobs:
     - uses: actions/setup-python@v1
       name: Install Python ${{ matrix.python_version }}
       with:
-        python-version: "3.7"
+        python-version: "3.9"
     - name: Install Python dependencies
       run: |
-        python -m pip install numpy pytest
+        python -m pip install numpy pytest cmake
     - name: Install dependencies on Ubuntu
       if: matrix.os == 'ubuntu-latest'
       run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/build-ale.yml
```diff
@@ -53,8 +53,11 @@ jobs:
...
       run: |
         mkdir build; cd build
-        cmake ../ -DUSE_SDL=ON
+        cmake ../ -DUSE_SDL=ON -DPython3_FIND_FRAMEWORK=NEVER
     - name: Build & Test ale
       run: |
         cd build
-        cmake --build . --config Release --target tests
+        cmake --build . --config Debug
+        ctest -C Debug --progress -VV
```
JesseFarebro, line 61: There's a known bug with VSCode that prevents us from using `cmake --build . --target test` so we'll use `ctest` for the time being.

.github/workflows/build-ale.yml
```diff
@@ -53,8 +53,11 @@ jobs:
       if: matrix.os != 'windows-latest'
       run: |
         mkdir build; cd build
-        cmake ../ -DUSE_SDL=ON
+        cmake ../ -DUSE_SDL=ON -DPython3_FIND_FRAMEWORK=NEVER
```
JesseFarebro, line 56: New CMake policy seems to find Python 3.9 framework ahead of the installed 3.9 by Github actions, disable finding the Python framework.

PR #377, opened 2020-12-05
Added setRAM function.

benblack769:
Upstreaming a PR from our project: https://github.com/PettingZoo-Team/Multi-Agent-ALE/pull/8. Similar PRs have been proposed before:  mgbellemare#247, mgbellemare#232,  but were not merged due to inactivity. 

Seems to be about time this change is actually merged with the main project.


justinkterry:
@JesseFarebro just following up on this

JesseFarebro:
I think adding `setRAM` is valuable. I've had to implement the same functionality myself for projects in the past. My main concern about `setRAM` was modifying the system data bus. The ALE never controlled for this and we actually corrupt the data bus when we call `processRAM`. It's probably not worth it to implement the proper guards around the data bus at this point. By doing so we'd change how certain collisions work in the ALE which would alter some games.

PR #382, opened 2020-12-31
ale-py: Gym registration + ROM management

JesseFarebro:
## Gym Registration

We'll now be able to use upstream versions of the ALE with Gym and we no longer have to rely on `atari-py`. This will function as follows, first install `gym` and `ale-py`: `pip install ale-py gym`. Then,

```py
import gym

# Version suffixes
gym.make("ale-py:Freeway-v0") # v0-versioned environments, stochastic frame-skip between (2, 5) uniformly.
gym.make("ale-py:Pong-v4")    # v4-versioned environments, sticky actions (p=1/4) instead of stochastic frame-skip

# Config suffixes
gym.make("ale-py:SpaceInvadersDeterministic-v0") # Fixed frame-skip, this overrides v0 stochastic frameskip.
gym.make("ale-py:HeroNoFrameskip-v4")            # No frameskip, i.e., frameskip of 1. This overrides v0 stochastic frameskip.
```

You can also specify a certain subset of environment parameters to `gym.make`. One notable addition over is the `render_mode` parameter. This follows the paradigm of [gym3](https://github.com/openai/gym3) which specifies two render modes, `human` and `rgb_array`.

If it's set to `human` we will render interactively using SDL at the ROM's prescribed framerate. On the other hand `rgb_array` will include a new key `rgb_array` in each steps metadata. So we have:

```py
env = gym.make("ale-py:Freeway-v4", render_mode="human") # This will render interactively using SDL

env = gym.make("ale-py:Freeway-v4", render_mode="rgb_array")
obs = env.reset()
_, _, _, metadata = env.step(0)
assert 'rgb_array' in metadata
```

This means that `env.render()` will be deprecated in the future. For the time being it displays a deprecation error to maintain backwards compatability. The reason the ALE will deprecate `render` is that it becomes too cumbersome to handle `render` in the SDL case. By using the `render_mode` parameter we can also now guarantee that we can render on every frame instead of every `frame_skip` frames + some other nice features such as audio support.

## ROM Management

We need a good way to handle importing ROMs to provide a clear upgrade path for Gym where ROMs are included by default. Currently, we'll provide two methods to import ROMs.

### `ale-import-roms`

We now expose a CLI interface to import ROMs. It accepts a single argument, a directory, which we'll recursively scan and import all valid ROMs. The usage looks as follows:

```sh
 # Error, Freeway not found
python -c "import gym; gym.make('ale-py:Freeway-v4')"

# Directory structure:
#     ./myroms/freeway.bin
ale-import-roms myroms

# Success!
python -c "import gym; gym.make('ale-py:Freeway-v4')"
```

### `ale-py.roms` plugins

Python packages can expose entry points via setuptools. We will scan all installed packages and look for any entry points that match `ale-py.roms`. If we find such an entry point we'll load the function that's been exposed and we expect it to return a list of filenames of the form `List[pathlib.Path]`.

Setting up a package to distribute ROMs should be as easy as registering the entry point as follows:

__setup.cfg__
```conf
[metadata]
name = atari57
version = 1.0.0

[options]
packages = find:
install_requires =
    ale-py

[options.package_data]
* = *.bin

[options.entry_points]
ale-py.roms =
  atari57 = atari57:import_atari57_roms
```

## Miscellaneous Changes

* We now return `std::optional<std::string>` with `isSupportedROM` which gives us a nicer API which I leverage during ROM importing.

* Gym `obs_type` now accepts `rgb`, `grayscale`, `ram`, where before it only accepted `image`, `ram`. I hope to add native C++ Area resizing and frame-stacking so we can do this all in the ALE which will hopefully give a noticeable speedup.

@christopherhesse


benblack769:
If you want to do native image resizing, note that cv2's implementation which is broadly used in RL is super highly optimized intrinsic code, and is way faster than any standards compliant c++ code out there. If you want to see what this looks like, check out this project, https://github.com/ethereon/lycon, which basically extracted out all the image resizing code from opencv. I would not recommend doing this in a single project.

PR #388, opened 2021-01-01
Remove `os_dependent` use C++17 FS

JesseFarebro:
The only reason we're still using `os_dependent` from Stella is for some basic filesystem checks. We can now use the standardized C++17 filesystem API. This PR also includes some cleanups to OSystem to remove any references to Stella's main loop and timing info.


src/os_dependent/SettingsWin32.cxx
```diff
@@ -1,36 +0,0 @@
...
-#include "emucore/bspf/bspf.hxx"
-#include "emucore/Settings.hxx"
-#include "os_dependent/SettingsWin32.hxx"
-
-// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-SettingsWin32::SettingsWin32(OSystem* osystem)
-  : Settings(osystem)
-{
-  // Anything less than this usually causes sound skipping
-  setInternal("fragsize", "2048");
```
JesseFarebro, line 28: This doesn't seem to be an issue.

Issue #366, opened 2020-09-07
Modes backwards compatibility. 


weepingwillowben:
Modes in ALE have been changing with [this](https://github.com/mgbellemare/Arcade-Learning-Environment/commit/d3f2b258ed4890a8e10a6ff876436f585d6e6cef) commit, and soon afterwards PR #365. 

Me and @justinkterry  are hoping to document interesting game modes for our Multi-Agent ALE project and were wondering how these modes might change moving forward. In particular:

1. Are modes in old games (like surround) untouchable for backwards compatibility reasons, even when they are hacked to start at 0?
2. If so, when will modes for new games (like othello) be locked in place?

justinkterry:
@tkoeppe is this something you have any insight into?

kieranhj:
One challenge is that for many ROMs the supported game modes are enumerated from 0, whereas the _actual_ modes  start at 1, at least from the user's point-of-view / as documented in the ROM manuals. This pattern was generally carried forward when the newer games were added. We could document the discrepancies to get an idea of how big a change that might be.

Having the game modes be consistent with the documentation certainly reduces confusion when attempting to articulate the differences in game play.

JesseFarebro:
@kieranhj do you have an idea of how much work it would be to get every mode to be consistent with their respective manual?

weepingwillowben:
@tkoeppe I also feel it would be far easier to use the library when the modes are consistent with their manual than if they start from 0 (as the mode byte defined by the ROM often does). The reasons are: 

1. This way, the Atari manuals can be used as documentation for ALE. 
2. People familiar with the Atari games, or who own consoles can use the modes much more easily. 

In terms of our particular vested interests in Multi-player ALE, the relevant games are Pong (Video Olympics) and Surround, both of which have several interesting multi-player modes we wish to document but the mode numberings for these games are currently not aligned with the manual. Surround maps manual mode 2 to ALE mode 0 and Pong maps manual mode 1 to ALE mode 0.

PR #397, opened 2021-01-02
changed mode numbers for pong and surround

benblack769:
This fixes the surround and pong game modes to align with the manual, as per discussion here #366.

### DIFF
--- a/src/games/supported/Pong.cpp
+++ b/src/games/supported/Pong.cpp
@@ -81,11 +81,7 @@ void PongSettings::loadState(Deserializer& ser) {
 
 // returns a list of mode that the game can be played in
 ModeVect PongSettings::getAvailableModes() {
-  ModeVect modes(getNumModes());
-  for (unsigned int i = 0; i < modes.size(); i++) {
-    modes[i] = i;
-  }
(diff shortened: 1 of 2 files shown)


JesseFarebro:
@weepingwillowben do you have any sense of how aligned other modes are with their manual, i.e., 1-indexed with the proper indices?

weepingwillowben:
@JesseFarebro 
My understanding is that most of the environments are currently 0 indexed, even though the manual is 1 indexed. Pong is an example of this. Pong is important to us because there are so many interesting multi-player modes in it. 

 A few were indexed in more bizarre manners, for example Surround's modes currently start at -1. Some of those were fixed by #365. 
 
 Then there are a few environments that do have modes in ALE, but not in the game manual, such as double dunk.

jkterry1:
@Sebastian-Griesbach

Would you be willing to create a new version of this PR with the merge conflicts fixed so it can be reviewed as part of our ongoing work to get the multi-agent ALE stuff merged in?

Sebastian-Griesbach:
@jkterry1 The updated PR is #712

PR #400, opened 2021-01-03
`Sound*` only requires access to `Settings` not `OSystem`

JesseFarebro:
We should only require what we use as a parameter. We don't require `OSystem` for anything other than `Settings`.


src/emucore/Console.hxx
```diff
@@ -19,6 +19,7 @@
 #ifndef CONSOLE_HXX
 #define CONSOLE_HXX
 
+class OSystem;
```
JesseFarebro, line 22: This forward declaration was coming from `SoundNull`.

PR #401, opened 2021-01-04
Move OSystem::Random -> System::Random for determinism

JesseFarebro:
The ALE is based on v2.4.2 of Stella. Shortly after the 2.4.2 release, the Stella authors moved the RNG from OSystem -> System. This was done to enable cartridges to randomize entries on **reset** instead of construction. This more accurately emulates the hardware and for our purposes allows us to more easily enable full determinism. We can do so by separating the emulator RNG and the environment RNG. Currently, we have two sources of emulator stochasticity

1) Cartridge RAM randomization
2) RAM I/O Timer

Currently, in the ALE, we were using the same RNG for emulator and environment randomness. This has the side effect that `setInt('random_seed', ...)` has a direct effect on some game's behaviour where they wouldn't be fully deterministic. We guarantee that games are deterministic so we should fix the emulator random seed and eventually get rid of these two sources of stochasticity.

With these changes, I thought we'd revisit the cloning API. The distinction between `cloneState` / `cloneSystemState`  has historically been confusing for new users. I propose that we eventually replace `cloneSystemState` by `cloneState(include_rng=True)`. We can include a default of `cloneState(include_rng=False)` for backwards compatibility. In this PR I've implemented the groundwork for `cloneState(include_rng)`. For the time being we've kept `cloneSystemState` / `restoreSystemState` for backwards compatibility. Furthermore, MGB proposed getting rid of the `load` / `save` mechanism in #178. This is now removed.

## Determinism Results

Thanks to @YuhangSong for posting his script in #246.

The following results test if the ALE is deterministic following `loadROM`, i.e.,

```
for each rom
    generate fixed trajectory of actions { A_0, ... A_99 }
    for i = 1 ... 400
        setInt('random_seed', i)
        loadROM( rom )
        replay fixed trajectory
        record observation
    end

    check if all observations are similar
end
```

### Before ~80% deterministic

<details><summary>Show Results</summary>

```
game:name_this_game grouped_num:1 distribution:[400]
game:asteroids grouped_num:1 distribution:[400]
game:chopper_command grouped_num:2 distribution:[383, 17]
game:pitfall2 grouped_num:1 distribution:[400]
game:space_war grouped_num:3 distribution:[201, 107, 92]
game:darkchambers grouped_num:1 distribution:[400]
game:video_cube grouped_num:1 distribution:[400]
game:superman grouped_num:1 distribution:[400]
game:atlantis2 grouped_num:1 distribution:[400]
game:pacman grouped_num:1 distribution:[400]
[...]

PR #413, opened 2021-04-26
Process the screen and RAM after restoring a state

anadrome:
This commit would make StellaEnvironment::restoreState() update the current screen and RAM to match the restored state. Currently, restoring a state and then calling getScreen() or getRAM() returns stale values until an action is taken, since only pressSelect() and emulate() update the m_screen and m_ram variables within StellaEnvironment. I don't think that behavior was intended, or at least it's not documented anywhere, and it's a bit confusing imo (I finally tracked down a bug in some of my own code to this behavior).

### DIFF
--- a/src/environment/stella_environment.cpp
+++ b/src/environment/stella_environment.cpp
@@ -128,6 +128,8 @@ ALEState StellaEnvironment::cloneState() {
 
 void StellaEnvironment::restoreState(const ALEState& target_state) {
   m_state.load(m_osystem, m_settings, m_cartridge_md5, target_state, false);
+  processScreen();
+  processRAM();
 }
 
 ALEState StellaEnvironment::cloneSystemState() {


JesseFarebro:
Hi @anadrome,

Thanks for taking the time to investigate this issue. Sadly I don't think the solution is as easy as calling `processRAM` + `processScreen`. A couple of problems, first is that `processRAM` mutates the state of the emulator via the system bus state. This can be overcome by adding a flag in `System::poke`, which protects the bus state from being mutated. The second is that `processScreen` won't do what you're expecting. `processScreen` simply copies over the current frame buffer. The problem being is that we don't restore the frame buffer and there's currently no way to re-generate that frame without mutating the emulator's state. As I'm sure the functionality you outline is desirable for many, I think there are a couple of solutions to these problems.

1. Serialize the RAM + frame buffer. We could add a flag to `cloneState`, which optionally serializes the RAM + frame buffer. This would be backwards compatible while allowing for a "full" restore, albeit with a larger serialized string.

2. The emulator is deterministic (with #401); it is possible to clone the state at time `t - 1` and on restore advance one frame, which would rehydrate the RAM + frame buffer. This has the advantage of maintaining the small checkpoint size we currently enjoy. The implementation might be a little tricky, and I'd like to write some functional tests if we went down this road. I have a test framework that hasn't been pushed, so this might be an option I explore in the future.

I hope this makes sense. I think the first option could be a good short-term solution if you're willing to go out of your way to implement it. If this is something, you don't have the time for, I'll add this to my TODO list for the summer. Thanks again for bringing this to our attention.

PR #427, opened 2021-09-16
Move ale-py.gym -> gym.envs.atari as a namespace package & hook into Gym plugin system

JesseFarebro:
As per the issue in https://github.com/openai/gym/issues/2407 raised by @lebrice there is some legacy code that checks the entry point of the Atari environment. This can actually be solved in an elegant way by taking advantage of how namespace packages are installed.

If we package `gym/envs/atari/*` while excluding `__init__.py` at `gym/` and `gym/envs/` BUT including `__init__.py` at `gym/envs/atari` our subpakage will get installed to Gym as we'd expect. This allows the ALE to remain in control over the environment code while not breaking the entry point of `gym.envs.atari:AtariEnv`.

Furthermore, with the new plugin architecture in https://github.com/openai/gym/pull/2383, https://github.com/openai/gym/pull/2409 we can now register our environments as well as the legacy environments using the plugin architecture. This has the advantage of auto registering the `ALE/*-v5` environments without needing to import `ale-py`.


tests/python/gym/test_gym_interface.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 pytest.importorskip("gym")
+pytest.importorskip("gym.envs.atari")
```
JesseFarebro, line 4: There's no way to hack PYTHONPATH where we could perform Gym tests without having to install `ale-py`. Skip these tests if we don't have `gym.envs.atari` installed.

PR #431, opened 2021-09-28
Add a missing std namespace qualifier

anadrome:
This causes a build failure for me when building without SDL support.


JesseFarebro:
Ahh, I actually ran into this and meant to fix it but it must have slipped my mind. Thanks, @anadrome!

PR #445, opened 2022-01-09
ALE Namespace wrapper for emucore

tuero:
The emucore code has naked class definitions such as `Device`, which makes it difficult to link with other libraries. One example of this is libtorch C++. This pull request wraps the code in emucore inside the ale namespace. The resulting effect is that this change allows to link against libtorch (which also defined a `Device`) without any linking errors.

### DIFF
--- a/src/ale_interface.cpp
+++ b/src/ale_interface.cpp
@@ -55,6 +55,7 @@
 namespace fs = std::filesystem;
 
 namespace ale {
+using namespace stella;
 
 // Display ALE welcome message
 std::string ALEInterface::welcomeMessage() {
--- a/src/ale_interface.hpp
+++ b/src/ale_interface.hpp
@@ -193,8 +193,8 @@ class ALEInterface {
   ScreenExporter* createScreenExporter(const std::string& path) const;
 
  public:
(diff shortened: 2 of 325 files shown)


JesseFarebro (changes requested):
Hi @tuero,

Thanks for the contribution! This has been on my to-do list for quite some time. The only thing I'd like to see changed is that everything in `emucore/` is namespaced as `ale::stella`. Other than that it looks good.

tuero:
@JesseFarebro Thanks for the suggestion! I've updated emucore to be under the `ale::stella` namespace. As a result, all references to any classes in emucore also need to reference the `stella` namespace (everything is already under `ale` but not `stella`). 

Unfortunately, although these changes are minor, a lot of files are touched as a result. I've broken it up into several commits grouped so that the pull request review is easier to follow. Since all commits are needed without breaking anything, I would suggest squashing them before this pull request gets merged

JesseFarebro:
@tuero sorry for the delay, I will be getting around to this. Thanks for making these changes, I just need to dedicate some time to look everything over before the next release. This will coincide with the release of Gym 0.22.

JesseFarebro:
Thanks for the contribution @tuero!

PR #457, opened 2022-05-13
Add `const` keyword to getters

AlessioZanga:
As the title says, this patch add missing const correctness to getters of the `ALEInterface` class.


JesseFarebro:
Hi @AlessioZanga, sorry for taking so long to get around to this PR. Thanks for the contribution! 🎉

Issue #475, opened 2022-10-31
Recording video no longer works when using Gym with ALE


SebastianHurubaru:
Hello,

since the change in Gym 0.26.0 of the render_mode flag, the gym.wrappers.RecordVideo does not work anymore with ALE!

When recording the video we are getting the following log message:

.../lib/python3.8/site-packages/gym/wrappers/monitoring/video_recorder.py:59: UserWarning: WARN: Disabling video recorder because environment <OrderEnforcing<PassiveEnvChecker<AtariEnv<ALE/Pong-v5>>>> was not initialized with any compatible video mode between rgb_array and rgb_array_list

We are setting the render_mode when creating the environment, as showed in the [blog](https://younis.dev/blog/render-api/) for the Gym 0.26.0 release, but the render_mode flag doesn't get set in the `AtariEnv`, as the class uses an internal variable `self._render_mode` instead, leading to the warning and videos not being recorded.

Here is a code sample that can help with the issue replication:

`
import gym

env = gym.make("ALE/Pong-v5", render_mode="rgb_array")

env = gym.wrappers.RecordVideo(
  env,
  video_folder='video',
  step_trigger=lambda x: x % 100 == 0
)

env.reset()
for t in range(100):
  action = env.action_space.sample()
  observation, reward, terminated, truncated, info = env.step(action)

env.close()
`

We are using Python 3.8.10 with Gym 0.26.2 and ale-py 0.8.0

JesseFarebro:
The fix has now been merged and will be released with v0.8.1.

Issue #471, opened 2022-10-07
[BUG] Atari envs don't define `self.render_mode` for compatibility with gym 0.26


RedTachyon:
In gym 0.26, we use a standard approach of setting `self.render_mode` during initialization, which is then mostly for internal management by the env, but it is also used by some wrappers, e.g. [VideoRecorder](https://github.com/Farama-Foundation/Gymnasium/blob/main/gymnasium/wrappers/monitoring/video_recorder.py).

With the current version of ALE, this doesn't work. Consider this snippet:
```python
import gym
from gym.wrappers import RecordVideo

env = gym.make('ALE/MontezumaRevenge-v5', render_mode='rgb_array')

env.unwrapped.render_mode = env.unwrapped._render_mode

env = RecordVideo(env, video_folder='videos')
state, info = env.reset()
done = False
while not done:
    action = env.action_space.sample()
    state, reward, terminated, truncated, info = env.step(action)
    done = terminated or truncated
env.close()
```

This throws up a warning that recording is disabled because the environment has not been defined with an appropriate `render_mode`.

The reason is that ALE internally records the render mode as `self._render_mode`, so the wrapper doesn't pick it up. 

I'm pretty sure the only needed fix is replacing `_render_mode` with `render_mode`, I can contribute a PR if that'd be useful.

JesseFarebro:
This is fixed as of v0.8.1 for the legacy Gym environment.

PR #476, opened 2022-11-20
FIX: add render_mode attribute

younik:
This PR addresses https://github.com/mgbellemare/Arcade-Learning-Environment/issues/471 and https://github.com/mgbellemare/Arcade-Learning-Environment/issues/475 by adding the property render_mode to AtariEnv.

Quick test:
```python
import gym
from gym.wrappers import RecordVideo

env = RecordVideo(
    gym.make('ALE/Breakout-v5', render_mode="rgb_array"),
    video_folder='video'
)

env.reset()
for _ in range(100):
    env.step(env.action_space.sample())

env.close()
```

@JesseFarebro

### DIFF
--- a/src/python/env/gym.py
+++ b/src/python/env/gym.py
@@ -125,6 +125,7 @@ class AtariEnv(gym.Env, utils.EzPickle):
             max_num_frames_per_episode,
             render_mode,
         )
+
         # Initialize ALE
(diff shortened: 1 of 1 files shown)


JesseFarebro:
Thanks for fixing this! 🙌

PR #486, opened 2023-05-05
Add missing imports for cstdint

### LINKED ISSUE #485: Not building under arch
I believe we got a newer version of glibc last week (2.37-3), and it seems since that version, the package is not building anymore.

It is complaining immediatly about missing import for uint32_t:

```
[  0%] Building CXX object src/CMakeFiles/ale.dir/common/ColourPalette.cpp.o
In file included from /home/ben/aur/python-ale-py/src/Arcade-Learning-Environment-0.8.1/src/common/ColourPalette.cpp:16:
/home/ben/aur/python-ale-py/src/Arcade-Learning-Environment-0.8.1/src/common/ColourPalette.hpp:34:3: error: ‘uint32_t’ does not name a type
   34 |   uint32_t getRGB(int val) const;
      |   ^~~~~~~~
/home/ben/aur/python-ale-py/src/Arcade-Learning-Environment-0.8.1/src/common/ColourPalette.hpp:21:1: note: ‘uint32_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
   20 | #include <string>
  +++ |+#include <cstdint>
   21 |
```

Did I just messed-up a flag, or that's to be expected?

benallard:
Closes #485

### DIFF
--- a/src/emucore/Switches.cxx
+++ b/src/emucore/Switches.cxx
@@ -16,8 +16,6 @@
 // $Id: Switches.cxx,v 1.7 2007/01/01 18:04:50 stephena Exp $
 //============================================================================
 
-#include <cstdint>
-
 #include "emucore/Event.hxx"
 #include "emucore/Props.hxx"
 #include "emucore/Switches.hxx"
--- a/src/emucore/Switches.cxx
+++ b/src/emucore/Switches.cxx
@@ -16,6 +16,8 @@
 // $Id: Switches.cxx,v 1.7 2007/01/01 18:04:50 stephena Exp $
 //============================================================================
 
+#include <cstdint>
+
 #include "emucore/Event.hxx"
 #include "emucore/Props.hxx"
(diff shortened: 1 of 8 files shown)


src/emucore/Switches.cxx
```diff
@@ -16,6 +16,8 @@
 // $Id: Switches.cxx,v 1.7 2007/01/01 18:04:50 stephena Exp $
 //============================================================================
 
+#include <cstdint>
```
JesseFarebro, line 19: I believe `Switches.hxx` also uses `uint8_t`. Could you move this import into `Switches.hxx` instead?

JesseFarebro (changes requested):
Thanks for these changes @benallard! Not sure how these were missed.

src/emucore/Switches.cxx
benallard: Sure, you're correct, done.

PR #496, opened 2023-08-23
Add website

pseudo-rnd-thoughts:
The initial setup for a website using the same style as all other farama websites

Todo
- [ ] Autobuild the environment documentation
- [x] Read the new documentation to check it is consistent
- [x] Fix the environment images / gifs
- [ ] Auto generate the gifs
- [x] Add github action 

I'm not going to implement the first or fourth items as the rooms don't work currently within the CI I believe


pseudo-rnd-thoughts:
@JesseFarebro Could you have a quick look, you should be able to preview the website by installing the requirements from `docs/requirements.txt` then running `make html` in docs

.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,50 @@
+name: Build main branch documentation website
+
+on:
+  push:
+    branches: [main]
```
mgoulao, line 5: This repo uses `master` instead of `main`

pseudo-rnd-thoughts: Thanks

jjshoots (approved):
LGTM.

PR #503, opened 2024-01-15
build: fix compilation with gcc 13

### LINKED ISSUE #502: Compilation fails with GCC 13
When trying to install `ale-py` with GCC13 as a compiler, the build fails with several instances of such error:
```
note: ‘uint8_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
```

I guess that `<cstdint>` has now to be explicitly included: https://gcc.gnu.org/gcc-13/porting_to.html

GaetanLepage:
This include is necessary for the project to compile with GCC 13.
See https://gcc.gnu.org/gcc-13/porting_to.html

Fixes #502

### DIFF
--- a/src/emucore/TIASnd.hxx
+++ b/src/emucore/TIASnd.hxx
@@ -18,6 +18,7 @@
 
 #ifndef TIASOUND_HXX
 #define TIASOUND_HXX
+#include <cstdint>
 
 namespace ale {
 namespace stella {


GaetanLepage:
From my testing, this is working.

pseudo-rnd-thoughts (approved):

PR #513, opened 2024-02-23
Includes roms within build for pypi and testing

jjshoots:
Changes:
1. Move roms installation from AutoROM to here, installation is via `ale-accept-license-install-roms`.
3. Some other minor QoL stuff.


JesseFarebro:
Not a huge fan of downloading ROMs from a gist, if y'all are going to install ROMs can't you reuse the AutoROM infrastructure and make it optional with a similar `accept-rom-license` extra.

pseudo-rnd-thoughts:
> Not a huge fan of downloading ROMs from a gist, if y'all are going to install ROMs can't you reuse the AutoROM infrastructure and make it optional with a similar accept-rom-license extra.

The AutoROM infrastructure is just downloading from the gist and unpacking from my knowledge

On the accept-rom-license, to my knowledge, we never explicitly explain what rom license people are accepting. 
Could we have a `ale-py` (without the roms) and an ale-py with the roms (using a --accept-rom-license) requirement?

src/python/__init__.py
```diff
@@ -65,5 +65,6 @@
 
     register_v0_v4_envs()
     register_v5_envs()
-except ImportError:
+except ImportError as e:
+    print(e)
```
pseudo-rnd-thoughts, line 69: Remove, currently, I would have it silently error if `Gymnasium` hasn't been installed otherwise a more helpful error rather than a weird printed exception when loading ale-py

src/python/registration.py
```diff
@@ -48,7 +60,7 @@ def _register_rom_configs(
 
                     # Register the environment
                     gymnasium.register(
-                        id=f"{prefix}{name}{flavour.suffix}-{config.version}",
+                        id=f"ALE/{prefix}{name}{flavour.suffix}-{config.version}",
```
pseudo-rnd-thoughts: For v0 and v4 environments, there is no ALE prefix while v5 should have ALE prefix
Does this not add for all?

src/python/roms/__init__.py
```diff
@@ -1,141 +1,90 @@
...
-        # If the ROMs resolved differently
-        if resolved_id != rom_id:
-            warnings.warn(
-                f"{package} contains ROM {rom_id} which doesn't resolve to {resolved_id}. "
-                "This is most likely caused by a filename mismatch.",
-            )
+def _download_roms():
+    """Downloads all roms in the form of a base64 file, then decodes it into a tar.gz, then extracts all the roms by matching it to the expected md5."""
+    # this is a map of {rom.bin : md5 checksum}
+    all_roms = json.load(open(Path(__file__).parent / "md5.json"))
```
pseudo-rnd-thoughts, line 16: Use `with open()` such that the file is closed correctly

src/python/roms/__init__.py
```diff
@@ -1,141 +1,90 @@
...
+
+        # get hash
+        md5 = hashlib.md5()
+        md5.update(rom_bytes)
+        md5_hash = md5.hexdigest()
+
+        # assert that the hash matches
+        assert (
+            md5_hash == all_roms[rom_name]
+        ), f"Rom {rom_name}'s hash does not match what was expected. Please report this to a dev."
```
pseudo-rnd-thoughts, line 56: Do you want to include the expected and actual hashes for debugging?

src/python/roms/__init__.py
```diff
@@ -1,141 +1,90 @@
...
+
+        # save this rom
+        rom_path = Path(__file__).parent / rom_name
+        open(rom_path, "wb").write(rom_bytes)
+
+        print(f"Downloaded and unpacked {rom_name}.")
```
pseudo-rnd-thoughts, line 62: Remove as this will be printed for each rom

pseudo-rnd-thoughts (changes requested):
Can we have the script run during build? 
It would be easier for the PyPI to include the ROMs directly rather than every user download but might not be possible with `accept-rom-license`

tests/python/test_atari_env.py
```diff
@@ -311,7 +311,7 @@ def test_terminal_signal(tetris_env):
 
 
 def test_render_exception(tetris_env):
-    with pytest.raises(TypeError):
+    with pytest.raises(gymnasium.error.ResetNeeded):
```
pseudo-rnd-thoughts, line 314: Could you add an `tetris_env.reset()` beforehand such that we have the same error

pseudo-rnd-thoughts (changes requested):
Could we run `check_env` for all roms during tests ?

src/python/roms/__init__.py
jjshoots: Don't think this is needed, since we pass the fp directly into json and the `load` function closes after reading the files, garbage collector should automatically close it for us.

src/python/roms/__init__.py
jjshoots: Probably not, I doubt that's very useful

src/python/roms/__init__.py
pseudo-rnd-thoughts: Just to make sense it is correct, can you change

JesseFarebro:
> Can we have the script run during build? It would be easier for the PyPI to include the ROMs directly rather than every user download but might not be possible with `accept-rom-license`

If you are going to package ROMs without any type of user intervention I'd prefer this than downloading them.

> @JesseFarebro That was in the older version of AutoROM, but it was causing a lot of problems for people who were on e.g.: university networks.

I see!

pseudo-rnd-thoughts:
@jjshoots It would be nicer to include the ROMs within the PyPI release as users wouldn't need to download anything else
Like the stella message when the first environment is created, could we do something similar for loading roms?

jjshoots:
@pseudo-rnd-thoughts That was the original approach (where users didn't need to do any external action).

Should I just revert to that?

We can move the license agreement one stage earlier to "We assume you will be supplying your own roms or have already accepted the license agreement prior to installing this library", instead of having an explicit `pip3 install ale-py[accept-rom-license]` which I still don't think is the right way to do this.

pseudo-rnd-thoughts:
Sorry for chopping and changing so much.

Yes, I think in the end this is the right approach with just a warning to the user about this.
Probably with a link to the website

jjshoots:
@pseudo-rnd-thoughts No worries! I'll make the changes and add in a warning with a small sleep.

src/python/roms/__init__.py
```diff
@@ -1,141 +1,104 @@
...
+    bin_file = f"{name}.bin"
+    bin_path = Path(__file__).parent / bin_file
 
+    # check if it exists within the the hash dictionary
+    all_roms = _get_all_rom_hashes()
+    if bin_file not in all_roms:
+        warnings.warn(f"Rom {name} not supported.")
+        return None
+
+    # if the bin_path doesn't exist, we assume the user already has the license and just download the roms
```
pseudo-rnd-thoughts, line 88: This shouldn't happen as the roms are provided within the build

src/python/roms/__init__.py
```diff
@@ -1,141 +1,104 @@
...
-            warnings.warn(
-                f"{package} contains unsupported ROM: {rom_path}",
-                category=ImportWarning,
-                stacklevel=2,
+from pathlib import Path
+
+import requests
+
+
+@functools.lru_cache(maxsize=1)
```
pseudo-rnd-thoughts, line 15: What does lru_cache do and why are we using it?

pseudo-rnd-thoughts (changes requested):
Otherwise, looks good to merge

src/python/roms/__init__.py
jjshoots: This just preloads the `rom.bin:md5` json file so we don't perform a disk read everytime we need to use the dict, only on the first call. Optionally we can just make it a raw globally scoped variable if preferred.

jjshoots:
@pseudo-rnd-thoughts yes, can confirm that is the approach. To me, that feels the same as including the ROMs within the PyPI package but with slightly less moving parts. Are you suggesting we have CI that downloads the ROMs and prepackages before publishing to PyPI instead? This approach may complicate things for dev since ROM download during dev must be handled externally then. This will also end up putting the ROMs directly within the tar.gz file on PyPI (not that anyone will look at that). Open to suggestions.

src/python/roms/__init__.py
pseudo-rnd-thoughts: Cool, that makes sense

jjshoots:
Discussed with @pseudo-rnd-thoughts on Discord, we agreed that packaging the roms as a tar.gz file within the PyPI package and unpacking after is the most ideal way to skirt around this grey area.

.gitignore
pseudo-rnd-thoughts, line 1: Do we need all of these new gitignores?

curl_tar_gz_b64.sh
pseudo-rnd-thoughts, line 1: move to .github

pyproject.toml
```diff
@@ -33,6 +33,7 @@ classifiers = [
 ]
 dependencies = [
   "numpy",
+  "requests",
```
pseudo-rnd-thoughts, line 36: remove?

src/python/roms/__init__.py
```diff
@@ -1,141 +1,101 @@
...
+
+def _unpack_roms() -> None:
+    """Unpacks all roms from the tar.gz file, then matches it to the expected md5 checksum."""
+    all_roms = _get_all_rom_hashes()
+
+    # load the b64 file
+    if (Path(__file__).parent / "Roms.tar.gz.b64").exists:
+        with open(Path(__file__).parent / "Roms.tar.gz.b64") as f:
+            tar_gz_b64 = f.read()
+    else:
```
pseudo-rnd-thoughts, line 29: Remove

src/python/roms/__init__.py
```diff
@@ -1,141 +1,101 @@
...
+    with open(Path(__file__).parent / "md5.json") as f:
+        return json.load(f)
+
+
+def _unpack_roms() -> None:
+    """Unpacks all roms from the tar.gz file, then matches it to the expected md5 checksum."""
+    all_roms = _get_all_rom_hashes()
+
+    # load the b64 file
+    if (Path(__file__).parent / "Roms.tar.gz.b64").exists:
```
pseudo-rnd-thoughts, line 26: Shouldn't this only be done if the tar hasn't been unzipped yet?

pseudo-rnd-thoughts (changes requested):
I think we need to rewrite the tests after this but I feel we are close to a release

.gitignore
jjshoots: It's a boilerplate python project gitignore, available [here](https://github.com/github/gitignore/blob/main/Python.gitignore). I think it's good practice to just include them.

curl_tar_gz_b64.sh
jjshoots: That'll make it slightly more obtuse for other devs to use I think.

src/python/roms/__init__.py
jjshoots: We aren't shipping the tar.gz, we're shipping the b64, so we need to decode -> unpack -> check sums.

curl_tar_gz_b64.sh
pseudo-rnd-thoughts: It should only be used by GitHub and makes it a bit more hidden for users who shouldn't be using it

src/python/roms/__init__.py
pseudo-rnd-thoughts: Does the unzip remove the file?

.gitignore
pseudo-rnd-thoughts: For the simplicity of the pr, can we remove and put in another pr?

src/python/roms/__init__.py
jjshoots: Nope, I fixed this so gonna resolve this.

.gitignore
jjshoots: I think the ship has sailed beyond that point, I don't see how having it here affects things honestly

jjshoots:
Not sure why 3.8 is failing btw

pseudo-rnd-thoughts (approved):

PR #516, opened 2024-03-27
Pull Gym env in favor of Gymnasium and include ROMs with PyPI

pseudo-rnd-thoughts:
With Gymnasium v1.0 being released soon, one of the features removed is the namespace entry point used to register environments behind the scene, through what is a hack imho. 
In short, previous with `ale-py` (and `shimmy`) installed, the following code is possible
```python
import gymnasium as gym

env = gym.make("ALE/Pong-v5")
```
That despite `ale-py` not being installed, then behind the scenes, Atari environments are still registered. 

While an interesting approach, the more this has been used, the more of a headache it has been found to be. 
Therefore, it was decided to be removed in v1.0

This means that future users would be required to do
```python
import gymnasium as gym
import ale_py

env = gym.make("ALE/Pong-v5")
```
explicitly, importing ale_py for the modules to be loaded

As a result, this PR makes the following changes
* remove gym env
* Add gymnasium env 
* Includes AutoROM roms within `pypi` (but not github) 
* Add python 3.12 support


pseudo-rnd-thoughts:
@JesseFarebro Hey, this is holding up Gymnasium's alpha 2, could you review this as soon as possible. Otherwise, I'll need to merge in the next week or so

.github/workflows/ci.yml
```diff
@@ -68,21 +62,18 @@ jobs:
...
-          - runs-on: windows-latest
-            python: '3.9'
-            triplet: x64-windows
-          - runs-on: windows-latest
-            python: '3.10'
-            triplet: x64-windows
-          - runs-on: windows-latest
-            python: '3.11'
-            triplet: x64-windows
+#          - runs-on: windows-latest
```
JesseFarebro, line 65: Any reason CI is disabled for Windows?

.github/workflows/release.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     - name: Unzip wheels asset
       run: |
         unzip wheels.zip -d dist
+    - name: Download tar.gz.b64
```
JesseFarebro, line 21: Could use a more descriptive name.

.pre-commit-config.yaml
```diff
@@ -1,12 +1,61 @@
...
+#    exclude: ^(src/common)|(src/emucore)|(src/environment)|(src/games)
+#    hooks:
+#      - id: codespell
+#        args:
+#          - --ignore-words-list=
+  - repo: https://github.com/PyCQA/flake8
+    rev: 7.0.0
     hooks:
-    -   id: black
+      - id: flake8
```
JesseFarebro, line 30: Should switch to ruff?

.pre-commit-config.yaml
```diff
@@ -1,12 +1,61 @@
...
+          - --max-complexity=30
+          - --max-line-length=456
+          - --show-source
+          - --statistics
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.15.0
+    hooks:
+      - id: pyupgrade
+        args: ["--py38-plus"]
+  - repo: https://github.com/PyCQA/isort
```
JesseFarebro, line 43: Should switch to ruff?

.pre-commit-config.yaml
```diff
@@ -1,12 +1,61 @@
...
-    -   id: black
+      - id: flake8
+        args:
+          - '--per-file-ignores=tests/python/test_atari_env.py:F811 tests/python/test_python_interface.py:F811'
+          - --ignore=E203,W503,E741
+          - --max-complexity=30
+          - --max-line-length=456
+          - --show-source
+          - --statistics
+  - repo: https://github.com/asottile/pyupgrade
```
JesseFarebro, line 38: Should switch to ruff?

.pre-commit-config.yaml
```diff
@@ -1,12 +1,61 @@
...
+    rev: v3.15.0
+    hooks:
+      - id: pyupgrade
+        args: ["--py38-plus"]
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+        args: ["--profile", "black"]
+  - repo: https://github.com/python/black
```
JesseFarebro, line 48: Should switch to ruff?

curl_tar_gz_b64.sh
JesseFarebro, line 1: Would be nice to have a more descriptive name. Also, since this is CI related would be nice to put it in .github/scripts or something.

md5.txt
JesseFarebro, line 1: Why is this being deleted? I would like to keep this as this is what we should abide by and I want to keep the history of this file for forensic purposes.

pyproject.toml
```diff
@@ -10,7 +10,7 @@ build-backend = "setuptools.build_meta"
 name = "ale-py"
 description = "The Arcade Learning Environment (ALE) - a platform for AI research."
 readme = "README.md"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
```
JesseFarebro, line 13: Maybe 3.9 as 3.8 is going to be EOL before the next major release?

pyproject.toml
```diff
@@ -42,30 +42,21 @@ dynamic = ["version"]
...
 packages = [
   "ale_py",
   "ale_py.roms",
-  "ale_py.env",
-  "ale_py.scripts"
 ]
-package-dir = {ale_py = "src/python", gym = "src/gym"}
-package-data = {"ale_py" = ["py.typed", "*.pyi", "**/*.pyi"], "ale_py.roms" = ["*.bin", "md5.txt"]}
+package-dir = {ale_py = "src/python"}
+package-data = {"ale_py" = ["py.typed", "*.pyi", "**/*.pyi"], "ale_py.roms" = ["md5.json", "Roms.tar.gz.b64"]}
```
JesseFarebro, line 59: Why include the tar.gz instead of just the bin files? I don't really like this.

JesseFarebro (changes requested):
Overall this looks fine, my biggest complaint is around getting rid of the ROM plugins. Just assuming that the ROMs will be packaged and this will work for everyone's environment isn't going to cut it. The ALE is widely used at many companies where this isn't true so having a more customizable way to load ROMs is required IMO. I don't see any reason why you can't do the following:

- Package the bin files during CD
- Use the ROM plugin infrastructure to load these ROMs from `ale_py.roms`, in fact, it already did this, there should be no need to change anything here. All that needed to be done is to extract the ROMs into `ale_py.roms` during the wheel build.

.github/workflows/release.yml
pseudo-rnd-thoughts: The CI is broken for an unknown reason, see https://github.com/Farama-Foundation/Arcade-Learning-Environment/issues/515 for more info

.pre-commit-config.yaml
pseudo-rnd-thoughts: We just copied in Gymansium's pre-commit, it certainly would be an option

md5.txt
pseudo-rnd-thoughts: @jjshoots Why did we change to `md5.txt` to `md5.json`?

pyproject.toml
pseudo-rnd-thoughts: @jjshoots You preferred to include the tar and not the bins, do you remember why?

src/python/env.py
```diff
@@ -151,19 +145,21 @@ def __init__(
...
+        self.load_game()
 
+        # initialize action space
         self._action_set = (
             self.ale.getLegalActionSet()
             if full_action_space
             else self.ale.getMinimalActionSet()
         )
-        self._action_space = spaces.Discrete(len(self._action_set))
+        self.action_space = spaces.Discrete(len(self._action_set))
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: As the class inherits from `gym.Env` which defines the `action_space` as an attribute and not a property then it feels strange to use a property here.

src/python/roms/__init__.py
```diff
@@ -1,141 +1,95 @@
...
-else:
-    import importlib_resources as resources
-# pylint: enable=g-import-not-at-top
-
-# Precedence is as follows:
-#  1. Internal ROMs
-#  2. External ROMs
-#  3. ROMs from atari-py.roms
-#  4. ROMs from atari-py-roms.roms
-_ROM_PLUGIN_REGISTRY: List[plugins.Plugin] = [
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Thanks for information, that is very helpful to know. We will look to see what we can do. 
If the ROMs are packaged inside the PyPI install, does that cause issues for google? 
The primarily issue I see is for users with additional external ROMs not included in AutoROMs standard set then I'm not sure how this would work with it.

pyproject.toml
jjshoots: @pseudo-rnd-thoughts Someone along the chain of command mentioned we cannot have the bins on the GH repo.

pseudo-rnd-thoughts:
Thanks for the review @JesseFarebro, I will talk to @jjshoots about the changes to the plugin system. 
It might be helpful to have a short discord call about this, if possible, to find to better route through. 
You can easily find me on the Gymnasium discord

The other suggested changes about Ruff, we can easily look at and could you look at the Windows CI issues linked above

jjshoots:
Thanks for the review @JesseFarebro. I'm addressing the inclusion of bins on CD at #520.

.github/workflows/ci.yml
pseudo-rnd-thoughts: See https://github.com/Farama-Foundation/Arcade-Learning-Environment/issues/515, we commented it out such that the CI passed in the meantime

src/python/roms/__init__.py
```diff
@@ -1,141 +1,95 @@
...
-                f"{package} contains unsupported ROM: {rom_path}",
-                category=ImportWarning,
-                stacklevel=2,
+from pathlib import Path
+
+
+@functools.lru_cache(maxsize=1)
+def _get_all_rom_hashes() -> dict[str, str]:
+    # this is a map of {rom.bin : md5 checksum}
+    with open(Path(__file__).parent / "md5.json") as f:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Could you elaborate "history purposes"? What history data is of interest?

src/python/roms/__init__.py
```diff
@@ -1,141 +1,95 @@
...
+        return json.load(f)
+
+
+def _unpack_roms() -> None:
```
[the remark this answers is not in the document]
pseudo-rnd-thoughts: Done

pseudo-rnd-thoughts:
@JesseFarebro Could you re-review? 

* changing to Ruff - We will investigate this but for simplicity, for this version, we will keep the updated pre-commit settings
* md5.txt - I don't understand the point about history, looking at the blame, there seem to be three different PRs that have updated it, none of them look critical to keep. Therefore, it seems to me reasonable to keep the updated JSON format. If I've missed something please say
*  plugin system - With the ROMs being included within the project downloads (except git clone), to my understanding, the only feature missing from the updated method is interoperability with atari-py roms. What feature in the plugin system that the updated version doesn't include is important for organisations, i.e., google. Otherwise, the updated version is much simpler for everyone, therefore, I would wish to remove it.

src/python/roms/__init__.py
```diff
@@ -1,141 +1,49 @@
...
 
-        # Failed to resolve ROM
-        if rom_path is None:
-            logging.debug(f"{package} did not resolve {rom_id}.")
-            continue
+def get_rom_path(name: str) -> Path | None:
+    """Expects name as a snake_case name, returns the full path of the .bin file if it's valid, otherwise returns None."""
+    # the theoretical location of the binary rom file
+    bin_file = f"{name}.bin"
+    bin_path = Path(__file__).parent / bin_file
```
JesseFarebro, line 27: Maybe prefer to use `importlib.resources` which will provide better compatibility.

JesseFarebro (commented):
Overall, looks good. Please consider my comment RE: ROM loading via an environment variable and if that's fixed I don't have any further comments.

---

> md5.txt - I don't understand the point about history, looking at the blame, there seem to be three different PRs that have updated it, none of them look critical to keep. Therefore, it seems to me reasonable to keep the updated JSON format. If I've missed something please say

It's fine, I can go hunting in the history to find any changes to `md5.txt`.

> plugin system - With the ROMs being included within the project downloads (except git clone), to my understanding, the only feature missing from the updated method is interoperability with atari-py roms. What feature in the plugin system that the updated version doesn't include is important for organisations, i.e., google. Otherwise, the updated version is much simpler for everyone, therefore, I would wish to remove it.

I took a look at the latest changes and this still wouldn't work for Google at least. At minimum, we would need a way to specify a different or additional directory to search for ROMs. If you really don't want to bring back the ROM plugins could you provide this functionality via an environment variable?

jjshoots:
@JesseFarebro Apologies for seemingly going in circles with this. What component of the plugin system is required at Google?

More specifically, the current release already includes the binaries as part of package data on PyPI. Running `pip3 install ale-py` would pull the binaries from the python index itself at build, it is no longer "lazy-downloaded" at runtime.

The only cases I can see where the plugin system is needed are:
1. Google scrubs all `.bin` file from packages downloaded via pip.
2. There are some atari games not within ale-py that Google is using, and requires external linking
3. There are modifications to existing atari bins that are not native to ale-py.

Either 3 of those are valid use cases, but it'd be helpful to know which one in specific Google (or other companies) are using so that the plugin system's inclusion is justified.

Once again, sorry if this is going around in circles, we just want to make sure that the package doesn't have code that non of the maintainers know why should exist.

JesseFarebro:
@jjshoots Google doesn't use pip, only the source code exists in our version control system. This is why we need a way at runtime to modify the path where the ROMs can be discovered. As I said in my previous comment, at minimum we need an additional search path for ROMs via an environment variable.

@pseudo-rnd-thoughts I'd probably try updating vcpkg to the latest version, they usually fix any major issues in a timely manner.

pseudo-rnd-thoughts:
@JesseFarebro that makes sense, we will add an environment variable for overriding the rom path 

@JesseFarebro i don't understand vcpkg enough, could you make a pr for it?

jjshoots:
We resorted to using an environment var instead of a kwarg. It's in #522

.github/workflows/ci.yml
pseudo-rnd-thoughts: Fixed in #523

pseudo-rnd-thoughts:
Thanks @JesseFarebro for your review, I'm going to merge this PR and cut a release in the next few days (hoping that the wheels ci works)

Tobias-Fischer:
Would you be ok to open a PR @pseudo-rnd-thoughts? Or even better take over maintenance here?

PR #520, opened 2024-04-10
Unpack ROMs to bins before building

jjshoots:
This PR changes the implementation such that ROMs are used in the build rather than the tar.gz as requested by @JesseFarebro in #516


jjshoots:
@pseudo-rnd-thoughts ready for review + merge

.github/workflows/ci.yml
```diff
@@ -105,7 +105,7 @@ jobs:
         doNotCache: true
 
     - name: Download tar.gz.b64
```
pseudo-rnd-thoughts, line 107: Jesse requested this name be improved

.github/workflows/release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       run: |
         unzip wheels.zip -d dist
     - name: Download tar.gz.b64
```
pseudo-rnd-thoughts, line 21: Same here

download_unpack_roms.sh
pseudo-rnd-thoughts, line 1: Jesse suggested that we move this to `.github` or `scripts`, basically out of root. 
We can add a note for developers about this for testing locally

src/python/roms/tetris.bin
pseudo-rnd-thoughts, line 1: Why has this changed?

pseudo-rnd-thoughts (changes requested):
Otherwise looks good

src/python/roms/tetris.bin
jjshoots: Python's unpack and bash's unpack produces fundamentally different sized files, but the checksums still pass: https://stackoverflow.com/questions/69585800/what-is-the-fundamental-difference-between-tar-unix-and-tarfile-python

jjshoots:
@pseudo-rnd-thoughts done.

jjshoots:
Ah shit something broke.

pseudo-rnd-thoughts (approved):

PR #521, opened 2024-04-23
V0.9.0 kwarg for rom dir

jjshoots:


jjshoots:
@pseudo-rnd-thoughts I think CI is failing for the same vcpkg thing.

jjshoots:
Closing because we should do an environment variable instead of a kwarg.

PR #523, opened 2024-04-23
Update CI (re-enable windows and pin macos)

pseudo-rnd-thoughts:
The windows and macos CI is failing, trying to update the CI versions to see if that does anything


pseudo-rnd-thoughts:
Previously, the CI used `macos-latest` which changed from `mac-12` to `macos-14` where 14 is M1. 
For simplicity, we revert to macos 12 with future PR required to update to M1

PR #529, opened 2024-05-21
ci: update macos runners

chenrui333:
test run, https://github.com/chenrui333/Arcade-Learning-Environment/actions/runs/9179613891


pseudo-rnd-thoughts:
The macos versions were done specifically chosen as we had issues with the others as this PR shows. 
Is there a reason to update them? Or can you understand the problem?

PR #533, opened 2024-06-19
Change project src folder layout to use subfolder 'ale'

### LINKED ISSUE #532: incorrect cmake configuration / include layout post installation
I have been working on packaging this project for later use in c++ with conan, the package manager. 
During this process I believe to have come across an inconsistency in the c++ layout of include headers:

In the source tree everything lives directly in `src`, i.e header files start directly at e.g. `src/ale_interface.hpp`. 
```
src
├── ale_interface.hpp
├── CMakeLists.txt
├── common
│   ├── ...
├── emucore
│   ├── ...
│   ├── OSystem.hxx
│   ├── ...
│   ├── Sound.hxx
│   ├── ...

```
The include directories are also mapped to `src` for the c++ targets in `src/CMakeLists.txt`:
```
include_directories(BEFORE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR})
```

However, for installation the layout is changed to a project-named one (which is correct to do) `{install_prefix}/include/ale`:
```
  # Install Header files
  install(DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/
          DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME}
          FILES_MATCHING
          REGEX "\.h((pp)?|(xx?))$"
          REGEX "os_dependent\/.*(Win32)\.(h|h(xx)?|h(pp)?|c(xx)?|c(pp)?)$" EXCLUDE)
```

The problem that this creates is the following now:

In `ale_interface.hpp`:
```
 include `emucore/OSystem.hxx`
```
and in `emucore/OSystem.hxx`:
```
#include "emucore/Sound.hxx"
```
This works in the src layout. `#include emucore/OSystem.hxx` in `ale_interface` will always work because `emucore` is found via a relative search from the interface file. 

But `#include "emucore/Sound.hxx"` cannot resolve `emucore` from a relative search, since the including file (`OSystem.hxx`) is inside the included folder (`emucore`), so it will need to be found relative to the include-directories (`src` at build-time). But at install-time these are now changed to `ale`.
```
include
└── ale
    ├── ale_interface.hpp
    ├── common
    │   ├── ...
    ├── emucore
    │   ├── ...
    │   ├── OSystem.hxx
    │   ├── ...
    │   ├── Sound.hxx

```

I see 2 ways to fix this:

1. Change the src layout to mimick the install layout, i.e. move all files into `src/ale` and prepend all includes for project files with `ale/`, e.g. `#include "ale/emucore/OSystem.hxx".
2. Ensure all relative includes in all files are not relative to the top-level include, but the file itself.

I believe option 1 is less work and can be achieved with the help of some regex search and an IDE. Option 2 is more work in case there a many files that have made this mistake.

I could provide a PR if you agree with this. Otherwise, let me know if I got something wrong :)

maichmueller:
Fixes #532


pseudo-rnd-thoughts:
@maichmueller I think the issue is related to NumPy 2.0, if you limit the `numpy<2.0` in `pyproject.toml` this will probably fix

maichmueller:
Hi @pseudo-rnd-thoughts ,

Thanks for the hint! But this small numpy change now seems to cause the… vcpkg step to fail? I don’t quite understand the CI setup in how the steps interoperate; the changes in commits often don’t appear to me to correlate with the errors in the CI steps.

pseudo-rnd-thoughts:
@maichmueller There were several issues with the CI that took a while to fix. 
Could you merge with main and see if vcpkg is fix

maichmueller:
@pseudo-rnd-thoughts ,

thanks for those fixes! It seems the underlying CI issue is then resolved and all tests run successfully. I am happy to bring this draft to full PR status then if you agree.

pseudo-rnd-thoughts:
@maichmueller Thanks for sorting this, I'm not confident enough to confirm implementation 
@JesseFarebro Could you look at this? I believe you should have the c++ expertise, if not please say

pseudo-rnd-thoughts (approved):
I think I understand now, looks good to me

Thanks again for the PR @maichmueller

PR #549, opened 2024-07-31
Continuous ALE

jjshoots:
# Todo:

- [x] Make Cpp API pure continuous
- [x] Python-side continuous, quantized, discrete


pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts:
could you update the project version, I think there are a couple of places this needs to be done including the ci.yml

pseudo-rnd-thoughts:
For testing, it might be worth reducing the number of tests for continuous down to only `{env-id}-v0` or v5, as we are basically doubling the testing time required currently

pseudo-rnd-thoughts:
We can do it in another PR

PR #552, opened 2024-08-10
Bump to 0.10.0

jjshoots:
Bump the version to 0.10.0


jjshoots:
@pseudo-rnd-thoughts The commit history is bodged because I did my fix before I merged, but we squash and merge later anyway so I don't think it matters.

I did a grep for 0.9.1 across the repo and these were the places where 0.9.1 showed up. Lmk if there are any more. Also I did 0.11.0 instead of 0.10.0 because I believe that 0.10.0 was for numpy 2.0?

We also need to add the changelog date at a later time.

pseudo-rnd-thoughts:
@jjshoots No worries, you updated the three places that are necessary
I made the NumPy 2.0 support release 0.9.1, so could you make this 0.10.0
Also, could you add 0.9.1 to the release notes

jjshoots:
@pseudo-rnd-thoughts Did the things. :)

pseudo-rnd-thoughts (approved):

PR #572, opened 2024-11-04
Add documentation for continuous action space

jjshoots:


pseudo-rnd-thoughts (changes requested):
Could you add similar details to https://github.com/Farama-Foundation/Arcade-Learning-Environment/blob/074037fc753e09620ef121e22dade5684fa75a78/docs/environments.md?plain=1#L165

Issue #575, opened 2024-11-14
GPL-3.0 license reference intended?


tapdiego-amzn:
Despite the project being licensed under GPL-2.0, the file linked below explicitly references GPL-3.0. Is this intended?

https://github.com/Farama-Foundation/Arcade-Learning-Environment/blob/6a7e0ae7d4212dc5fa074e6c9d6d8aad463e871a/src/ale/common/Constants.cpp#L4

pseudo-rnd-thoughts:
Thanks for raising that, looking at it, the license has always been GPL 2.0 
It seems almost from the beginning that this difference existed - https://github.com/Farama-Foundation/Arcade-Learning-Environment/commit/9779b965297bbc0b5d68bca92b221b1ac727411a
Here, the License.txt and Constants.cpp were uploaded with contradicting license names. 

@tapdiego-amzn Would you be interested in making a PR to update all gpl-3.0 references consistent and about gpl-2.0?

tapdiego-amzn:
Sure. The GPL reference linked above is the only one explicitly mentioning version 3 that I could find, so it shouldn't take long.

PR #576, opened 2024-11-15
Removing GPL-3.0 license reference

tapdiego-amzn:
Fixes issue #575.


pseudo-rnd-thoughts (approved):
Thanks for checking more generally

PR #579, opened 2024-11-25
Add support for audio queries

gkennickell:
Adds audio support queries based on https://github.com/Farama-Foundation/Arcade-Learning-Environment/pull/233 updated to work with latest. This is a simplified/refactored version, discussion in: https://github.com/Farama-Foundation/Arcade-Learning-Environment/issues/577.

The atari_env.py support in the initial commit is not final, stubbed in for feedback purposes.


pseudo-rnd-thoughts (changes requested):
I think looks good. It would be good to do some testing with an actual algorithm. 

For the python stage, it wouldn't be a "separate" return value, rather I think we should do a Dictionary of `{"image": image_obs, "sound": sound_obs}` if `sound_obs==True`. 

We will need to update the `observation_space` if we are doing that. I'm happy to make some of those changes if you are uncertain. 

@jjshoots Do you have any thoughts?

pseudo-rnd-thoughts:
> I'm not as familiar with the gym api, so would definitely appreciate help integrating the feature in the cleanest way. I've amended the commit to remove the draft portion of the changes to env.py.

No worries, I have added the relevant code.
Could you add the following code to the end of the `tests/python/test_atari_env.py`

```python
def test_sound_obs():
    env = gymnasium.make("ALE/MsPacman-v5", sound_obs=True)

    with warnings.catch_warnings(record=True) as caught_warnings:
        check_env(env.unwrapped, skip_render_check=True)

    assert caught_warnings == [], [caught.message.args[0] for caught in caught_warnings]
```

Another note is our pre-commit raised the following issue. Could you add this. 
```shell
check that executables have shebangs.....................................Failed
- hook id: check-executables-have-shebangs
- exit code: 1

src/ale/common/SoundRaw.cxx: marked executable but has no (or invalid) shebang!
  If it isn't supposed to be executable, try: `chmod -x src/ale/common/SoundRaw.cxx`
  If on Windows, you may also need to: `git add --chmod=-x src/ale/common/SoundRaw.cxx`
  If it is supposed to be executable, double-check its shebang.
src/ale/common/SoundRaw.hxx: marked executable but has no (or invalid) shebang!
  If it isn't supposed to be executable, try: `chmod -x src/ale/common/SoundRaw.hxx`
  If on Windows, you may also need to: `git add --chmod=-x src/ale/common/SoundRaw.hxx`
  If it is supposed to be executable, double-check its shebang.
```

Otherwise, I suspect that we might be close to ready to merge.

pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts (changes requested):
@gkennickell Sorry, one last change, could you add some documentation to `docs/cpp_interface.md` on accessing the audio queries.

Then we should be good to merge

gkennickell:
Added simple documentation in docs/cpp_interface.md + fixed pre-commit tests. Please let me know if more documentation is warranted (for instance, in environments.md) or minimal is a good start.
And, thanks for the help on pushing this through.

pseudo-rnd-thoughts (approved):
Looks good, I'll add more documentation later

PR #581, opened 2024-11-27
Add CALE instructions and citation to README

psc-g:
Adding instructions for how to use CALE (Continuous ALE), as well as citation instructions.

Will also add instructions to https://ale.farama.org/environments/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,11 @@ while not episode_over:
 env.close()
 ```
 
+To run with continuous actions, you can simply modify the call to `gym.make` above with:
+```python
+env = gym.make('ALE/Breakout-v5', continuous=True, render_mode="human")
+```
+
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks Pablo, that looks great

psc-g:
Thanks for the quick review, Mark!

It seems the changes are not yet reflected on the gymnasium atari page (https://ale.farama.org/environments/). Is there a manual step that needs to run, or will this eventually be automatically picked up?

pseudo-rnd-thoughts:
The website is version controlled, so it does exist for main, https://ale.farama.org/main/environments/#continuous-action-space
I will see if we can make a manual version control to update the website

psc-g:
ah perfect, thanks!

PR #582, opened 2024-11-28
Update to Macos-13 runner 

pseudo-rnd-thoughts:
* Update to Python 3.9+
* Update to use Macos 13 action runners as MacOS 12 is deprecated and no longer supported on 3rd Dec 2024
* Updated SDL2 to 2.24.2 as 2.0.16 have deprecated code that xcode 14 / 15 would raise an error (this is required due to macos 13) - https://github.com/libsdl-org/SDL/issues/5846

For future explorers, when vcpkg says `CMAKE_CXX_COMPILER` not found, this is not true, rather one of the packages has failed so actually print the subpackage install logs and see where one has an error. You will probably need to update your subpackage version or the upstream repo itself.

FYI https://github.com/actions/runner-images Macos 13 is the x86 runner and macos14 is the arm64 runner


pseudo-rnd-thoughts:
Lesson to learn, read the logs and the log files if given very carefullly, won't just scroll to the bottom

PR #585, opened 2024-12-20
"Noframeskip" should have "F" in capital in the main docs as it is case-sensitive and otherwise a crash occurs

RaghuSpaceRajan:


pseudo-rnd-thoughts (approved):
Thanks for spotting that

PR #591, opened 2025-02-06
Use system CMake if available

mgorny:
Instead of depending on `cmake` and `ninja` executables unconditionally, add these dependencies only if system executables are not available. This ensures that system CMake is used when available (instead of being overridden by the PyPI build of CMake), so the user can benefit from improved portability due to downstream patching and avoids unnecessary dependencies.

This also adjusts the generator choice to support system `ninja` install, i.e. one without a `ninja` module.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,12 +64,6 @@ jobs:
       with:
         python-version: ${{ matrix.python }}
 
-    - name: Install windows cmake
-      # TODO(jfarebro): There's a bug with Windows cmake and PEP517 builds via pip install.
-      # As a temporary workaround installing cmake outside of the isolated env seems to work.
-      run: python -m pip install --user cmake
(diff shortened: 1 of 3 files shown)


pseudo-rnd-thoughts:
Thanks for the PR @mgorny
I suspect the issue with windows is related to https://github.com/Farama-Foundation/Arcade-Learning-Environment/blob/6b57d4250dbcc2f3bed3d18ffefcc7a69c55c055/.github/workflows/ci.yml#L67
Any ideas?

mgorny:
Hmm, that's an interesting problem. FWICS that implies that no entry points installed via `pip install --user` can work when you're inside another virtualenv. My first idea would be to try removing the workaround and see if it works.

pseudo-rnd-thoughts (approved):
Thanks @mgorny, the changes look good

mgorny:
Thanks!

PR #599, opened 2025-03-25
Adds native vector implementation

pseudo-rnd-thoughts:
Projects like `SampleFactory` and `EnvPool` have provided C++ based vectorisators for ALE, significantly increasing the number of frames per second that can be sampled.

However, `EnvPool` is no longer maintained and to make it easier and compatible with Gymnasium's vector API, this PR adds a `EnvPool` based (though highly highly simplified) vector environment.

The implementation has four main parts
1.  `preprocessed_env.hpp` - Runs the standard preprocessing computation for Atari (frame stack, frame skip, maxpooling, resizing, etc)
2. `async_vectorizers.hpp` - Takes a list of preprocessed environments that are run asynchronously that we push actions to then return a collection of results once computed. 
3. `ale_vector_python_interface.cpp/hpp` - PyBind11 bindings to convert the vectorizer's list of environment data into a batched version that is easier to use by users
4. `vector_env.py` - This is a python class for easily using the vectorizer. 

Known Problems
* For MacOS ARM (not x86), the ALE vector and AtariPreprocess are NOT equivalent for all timesteps, however, the difference in only 1 or 2 pixels with a difference of 1. The issue is almost certainly with OpenCV with a different implement between MacOS ARM and x86.

Features not included in the PR (todo later)
* This PR only supports next-step autoreset mode, future PRs could look to add disabled autoreset mode (relatively easy) and same-step autoreset mode (more difficult)
* XLA support - The EnvPool implementation adds support for `jit` accelerating the environment, however, the jax foreign function interface has changed significantly since the EnvPool therefore will need to be implemented from scratch. 
* Testing - This PR is minimally tested, particularly checking that agents can be successfully trained


jjshoots:
Reviewing it soon, thanks for the work on this!

jjshoots (approved):
Admittedly, a lot of it flew over my head and I never ran the code. But thus doesn't seem to break any existing functionality so I thibk it's good to go. Any downstream bugs can be fixed as they come a la open source.

pseudo-rnd-thoughts:
@jjshoots Thanks for the review, it was complex code and took me a long time to implement so don't worry if you don't fully understand

PR #602, opened 2025-04-28
Update pacman.md

rstrauss127:
Cleaned up 'Action Space' table by removing repeated column headers then reordering the elements in the table.

### DIFF
--- a/docs/environments/pacman.md
+++ b/docs/environments/pacman.md
@@ -29,10 +29,13 @@ Pacman has the action space of `Discrete(5)` with the table below listing the me
 To enable all 18 possible actions that can be performed on an Atari 2600, specify `full_action_space=True` during
 initialization or by passing `full_action_space=True` to `gymnasium.make`.
 
-| Value   | Meaning   | Value   | Meaning   | Value   | Meaning   |
-|---------|-----------|---------|-----------|---------|-----------|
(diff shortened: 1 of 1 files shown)


pseudo-rnd-thoughts:
Thanks for the PR, however, for environments with more actions (e.g. https://ale.farama.org/environments/alien/) then having multiple columns makes more sense

pseudo-rnd-thoughts:
Would you have a solution for those environments? We would want a single action meaning table style for all environments

rstrauss127:
Hey thanks for the quick reply and additional information! I totally understand if the tables make sense the way the are! I appreiciate you're response. Below are my thoughts: 

1. I would recommend a flat Action table for all environments (how it appears on the [Environment Specifications page](https://ale.farama.org/env-spec/)). 
2. Additionally, the actions can be grouped by type (movement vs. fire). Though, this option might make it harder for a user to copy/paste the entire action space. 

Best Regards!

pseudo-rnd-thoughts:
I added this change for all environments in https://github.com/Farama-Foundation/Arcade-Learning-Environment/pull/561

You can see the changes in https://ale.farama.org/main/environments/adventure/

rstrauss127:
Awesome thanks!

PR #628, opened 2025-07-12
Add features x11 and wayland to sdl

### LINKED ISSUE #625: Visually not getting rendered in Ubuntu
Install using ` pip install ale-py`

The game seems to be running in the background as the environment state gets updated when checked using the python debugger, but no window appears for rendering.
I have check the installation of pygame seperately and it is working perfectly. All other compartments of gymnasium seems to be working fine.

System:
Ubuntu 24.04 NVDIA GTX1650 Inside a virtual envrioment.

Code:
```
import gymnasium as gym

import ale_py
gym.register_envs(ale_py)

# Initialise the environment
env = gym.make("ALE/Breakout-v5", render_mode="human")

# Reset the environment to generate the first observation
observation, info = env.reset(seed=42)
for _ in range(1000):
    # this is where you would insert your policy
    action = env.action_space.sample()

    # step (transition) through the environment with the action
    # receiving the next observation, reward and if the episode has terminated or truncated
    observation, reward, terminated, truncated, info = env.step(action)

    # If the episode has ended then we can reset to start a new episode
    if terminated or truncated:
        observation, info = env.reset()

env.close()
```

As i run this file, it print on the terminal this:

>  A.L.E: Arcade Learning Environment (version 0.11.1+2750686)
[Powered by Stella]

unexploredtest:
in commit fa48461, the sdl version was upgraded from `2.24.2` to `2.30.11` in vcpkg. From what I understood, in older versions even if you disable default features, SDL would be built with x11 anyways (?) however that changed since version `2.26.2` in vcpkg and not using default features resulted in neither building with x11 nor with wayland support, which results in having the no display issue. 

Here we explicitly specify to use features x11 and wayland, which fixes the issue. We don't enable default features as that includes `ibus` and `dbus` which aren't necessary and cause build issues anyways.

Fixes #625.

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -11,7 +11,16 @@
       "dependencies": [
         {
           "name": "sdl2",
-          "features": [],
+          "features": [
+            {
+              "name": "x11",
+              "platform": "linux"
+            },
+            {
+              "name": "wayland",
+              "platform": "linux"
+            }
+          ],
           "default-features": false
         }
       ]


unexploredtest:
Yeah will add shortly

unexploredtest:
@pseudo-rnd-thoughts Fix added

pseudo-rnd-thoughts (approved):

PR #638, opened 2025-10-18
Fix action index calculation for continuous actions

markub3327:
Expected Behavior:
When the joystick is centered, both X and Y should return 0, keeping the player stationary.

Actual Behavior:
Centered joystick returns -1 for both X and Y, moving the player down-left.

### DIFF
--- a/src/ale/python/vector_env.py
+++ b/src/ale/python/vector_env.py
@@ -198,12 +198,12 @@ class AtariVectorEnv(VectorEnv):
             y = actions[:, 0] * np.sin(actions[:, 1])
 
             horizontal = (
-                -(x < self.continuous_action_threshold).astype(np.int32)
+                -(x < -self.continuous_action_threshold).astype(np.int32)
                 + (x > self.continuous_action_threshold).astype(np.int32)
                 + 1
             )
             vertical = (
-                -(y < self.continuous_action_threshold).astype(np.int32)
+                -(y < -self.continuous_action_threshold).astype(np.int32)
                 + (y > self.continuous_action_threshold).astype(np.int32)
                 + 1
             )


pseudo-rnd-thoughts:
Thanks for this PR, could you create test to confirm the expected behaviour?

markub3327:
Hi @pseudo-rnd-thoughts

Of course. Here’s an example test using a PyGame joystick.
Please note that a joystick device needs to be connected to your computer.

Thanks!

```python
import numpy as np
import pygame
import sys
import gymnasium as gym
import ale_py
import math

gym.register_envs(ale_py)

# Initialize Pygame and joystick module
pygame.init()
pygame.joystick.init()

# Detect connected joysticks
joystick_count = pygame.joystick.get_count()
if joystick_count == 0:
    print("No joysticks connected.")
    sys.exit()

print(f"{joystick_count} joystick(s) detected.")

# Initialize all joysticks
joy = pygame.joystick.Joystick(0)
joy.init()
print(f"Initialized joystick: {joy.get_name()}")

env_name = "ALE/Pitfall-v5"
env = gym.make(env_name, render_mode="human", continuous=True)

a = np.zeros((env.action_space.shape[0],), dtype=np.float32)
print(f"Action space: {env.action_space}")

def register_input():
    global quit_game, a

    pygame.event.pump()  # Update joystick state

    # Process all events
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            pygame.quit()
            sys.exit()

    # Atari 2600
    x_axis, y_axis = joy.get_axis(0), joy.get_axis(1)
    a[0] = math.sqrt(x_axis ** 2 + y_axis ** 2)
    a[1] = math.atan2(y_axis, x_axis) # atan2 returns angle in range [-pi, pi]
    a[2] = joy.get_button(0)

    if joy.get_button(1) == 1:
        quit_game = True

try:
    quit_game = False
    while not quit_game:
        s, info = env.reset()
        total_reward = 0.0
        steps = 0
        episode_start = True
        while True:
            register_input()
            next_s, r, terminated, truncated, info = env.step(a)

            total_reward += r
            episode_start = False
            if steps % 200 == 0 or terminated or truncated:
                print("\naction  " + str([f"{x:+0.4f}" for x in a]))
                print(f"step {steps} total_reward {total_reward:+0.4f}")
            steps += 1
            s = next_s
            if terminated or truncated or quit_game:
                break

            pygame.time.wait(int(1000.0/30.0))
    env.close()

except KeyboardInterrupt:
    print("\nExiting.")
    pygame.quit()
    sys.exit()
```

pseudo-rnd-thoughts:
Sadly I don't have a joystick controller. 
@markub3327 Could you add a test to ensure that this doesn't happen again

Also I believe that the same bug exists in the `AtariVectorEnv` implementation which is why the test is failing. Could you see if you can fix this as well

PR #652, opened 2025-11-28
Minimise data copies in vectorizer

pseudo-rnd-thoughts:
The largest bottleneck in the vectorizer was the memory transfer, as the observations were copied three different times. 
This PR optimises the number of data copies in particular for the observation from three to one (the optimal).  

We achieve this through 
* preprocessing environments push their observations directly to the output array. 
* for the same step autoreset, rather than copy all observations to the final observation if any of the sub-environment's episodes are over, we only update for the environments whose episode is over.


pseudo-rnd-thoughts:
A future PR will resolve any performance issues

PR #664, opened 2025-12-31
Fix source_repository URL case sensitivity

Nafyaz:


pseudo-rnd-thoughts (approved):

pseudo-rnd-thoughts:
Thanks for spotting that

Issue #654, opened 2025-12-08
Pitfall does not terminate after timer runs out


Sebastian-Griesbach:
In the description of [Pitfall](https://ale.farama.org/environments/pitfall/) it says
> The game is over if you collect all the treasures or if you die or if the time runs out.

However the environment does not terminate when the timer runs out. The player is simply unable to move but the environment keeps running.
The commonly used truncation limit for training atari games of 30 minutes is longer than the 20 minute timer. Which can lead to a lot of wasted transitions in a state of the game where the agent can no longer perform any action if the timer ran out.
I could of course change the truncation to accommodate this, but the timer running out should be a termination not a truncation as the agent can no longer collect any rewards after.
Is there some easy way for me to fix this in a fork of ALE? I could create a wrapper that counts the frames and terminates after 20 minutes but I'd prefer to fix this in the specific game itself somehow.
In the long term It might be worth considering changing this in ALE. As termination on timeout seems to be the intended behavior.

Sebastian-Griesbach:
I found that some episodes do terminate at the time limit and some don't. I'm not yet sure when each behavior occurs.

pseudo-rnd-thoughts:
This is a historic problem https://github.com/Farama-Foundation/Arcade-Learning-Environment/issues/268
The problem is "fixing" this bug and a couple of others could / should invalidate comparisons between the "buggy" an "fixed" ALE versions. 
Therefore, I've been neverous to make these changes in order people can trust the project can provide fair comparisons. 

The fix would be to do a v1.0 and v6 of the environments along with through performance comparisons to check how impactful these changes are to performance and therefore comparisons to old versions

pseudo-rnd-thoughts:
Reproducibility is core to making science work and crucial for fair comparisons. 
I don't know what affect this will have to fix this bug, but imagine that this improves performance for most algorithms by 10% on pitfall, this can produce suddenly unfair comparisons to previous work. 
Therefore, yes, we can make a v6 of the environments with the environment fixes however we need to be explicit about what the changes are, the performance impact, etc. 

I hope this explains my nervousness

Issue #268, opened 2019-08-23
Pitfall game over check doesn't seem to work


christopherhesse:
Using the `pitfall.bin` specified by the md5 file (`3e90cf23106f2e08b2781e41299de556`) if you start a game and then do nothing for 20 minutes, the timer runs out but `game_over()` is not set to true.

It looks like there is a check in the code, but it does not seem to work for me: https://github.com/mgbellemare/Arcade-Learning-Environment/blob/master/src/games/supported/Pitfall.cpp#L58

The following script should eventually finish but it does not:

```
import atari_py

ale = atari_py.ALEInterface()
ale.setFloat("repeat_action_probability".encode("utf-8"), 0.0)
game_path = atari_py.get_game_path("pitfall")
ale.loadROM(game_path)
ale.reset_game()
step = 0
while True:
    if step % 1000 == 0:
        print(step)
    ale.act(0)
    step += 1
    if ale.game_over():
        print(f"done at step {step}")
        break
```

mgbellemare:
Thanks for flagging this! I'll take a look post-AAAI.

Profesor09:
One potential fix could be to manually check for the timer value in the while loop and set game_over() to true if the timer value exceeds a certain threshold.

Here's an example of what the updated while loop could look like:

python
Copy code
while True:
    if step % 1000 == 0:
        print(step)
    ale.act(0)
    step += 1
    if ale.getRAM()[17] >= 250: # check if timer value exceeds threshold
        ale.setEpisodeOver(True) # set episode over to true
    if ale.game_over():
        print(f"done at step {step}")
        break
This code snippet checks the RAM value at index 17 (which corresponds to the timer value) and sets the episode over to true if the timer value exceeds 250. This should trigger the game_over() function to be set to true and end the while loop.

Note that the threshold value of 250 may need to be adjusted depending on the specific game and its timer implementation.

PR #667, opened 2026-02-05
Adding timeout termination to PitFall

Sebastian-Griesbach:
It seems that the current code actually does not contain any intended detection for the Timeout.

`logo_timer` is non zero when the Activision Logo scrolls in the bottom left.

This is happening
1) At the beginning of the game before the player moved
2) When the player lost all lives
3) When the timer ran out 
Apparently the `logo_timer` was put here because the `lives_bytes` alone can temporarily take on a zero value during the death transition.

The current Termination condition `m_terminal` thus only checks if all lives have been lost. There is no detection for the timer reaching zero. I added two variables tracking minutes and seconds and added it to the termination condition.

I don't see how to reasonably add tests for this as the timer only exists
1) in ram what I am using now
2) Visually which is hard to use in a test.

I wrote a little script that records the last view frames of a episode to visually confirm that the timer runs out right before timeout termination which I can provide if needed.

Edit: I did not know that I can not add issue references in the title: #654, #268

### DIFF
--- a/src/ale/games/supported/Pitfall.cpp
+++ b/src/ale/games/supported/Pitfall.cpp
@@ -49,9 +49,14 @@ void PitfallSettings::step(const System& system) {
 
   // update terminal status
   int lives_byte = readRam(&system, 0x80) >> 4;
-  // The value at 09xE will be nonzero if we cannot control the player
+  // The value at 0x9E will be nonzero if we cannot control the player
   int logo_timer = readRam(&system, 0x9E);
-  m_terminal = lives_byte == 0 && logo_timer != 0;
+  // The game timer (displayed as MM:SS) is stored in BCD format at 0xD8 (minutes) and 0xD9 (seconds)
+  int timer_minutes = readRam(&system, 0xD8);
+  int timer_seconds = readRam(&system, 0xD9);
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (changes requested):

jkterry1:
The general rule with the ALE historically has always been to never alter ROM mechanics from decades before, the original ALE paper when into the reasoning behind this extensively

pseudo-rnd-thoughts:
> The general rule with the ALE historically has always been to never alter ROM mechanics from decades before, the original ALE paper when into the reasoning behind this extensively

I agree with @jkterry1 however this isn't modifying the ROM but how we "process" the game data. 
My suggestion would be able to have v6 version of some environments where we "fix" some game implementations where we "processed" the data wrong. i.e., ignored the timeout, overflow on rewards

Sebastian-Griesbach:
I reverted the PR to my first commit which did not contain any options and simply fixes the unintended behavior by extracting the timer from RAM. I think this is the simplest way of dealing with this. How would we make a v6 out of this?

jkterry1:
@Sebastian-Griesbach could you please rebase off main?

Sebastian-Griesbach:
To be transparent I am not fully confident with some git functionalities like rebase.
But looks like I did it right without breaking anything.

PR #669, opened 2026-02-12
Fix statebuffer race

josephdviviano:
The test failing in https://github.com/Farama-Foundation/Arcade-Learning-Environment/pull/668 is actually due to a flakey race condition that has failed in previous runs of CI - on master — on Windows, in CI run 20594583659 from December 30.

I think, in unordered mode, write_idx_.fetch_add(1) assigns a circular buffer position, but the actual data write happens after the position is claimed. Another thread can finish its write and increment count_ past the batch boundary, causing collect() to read a position that the first thread hasn't written to yet.

I was able to reproduce on my machine with the following test on the current codebase, but it should pass currently.

### DIFF
--- a/src/ale/vector/utils.hpp
+++ b/src/ale/vector/utils.hpp
@@ -142,33 +142,38 @@ namespace ale::vector {
               timesteps_(num_envs_),
               slot_ready_(num_envs_),
               count_(0),
-              sem_ready_(0),
-              sem_read_(1) {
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts:
#652 has been merged and has resolved this issue though further optimisation is warranted

josephdviviano:
That's interesting - the other PR seems quite unrelated


Joseph (Mobile)

PR #694, opened 2026-06-01
feat: multi-ROM support for vectorized environments

AdrianOrenstein:
Extends `AtariVectorEnv` to accept multiple ROMs, enabling parallel training across seeds and games.

Previously, all environments in `AtariVectorEnv` ran the same ROM. Now `rom_paths` accepts
a list, and `num_envs` replicates each ROM: `["pong", "breakout"]` with `num_envs=2` expands
to `["pong", "pong", "breakout", "breakout"]`.

**Renames:**
- `ALEVectorInterface.rom_path` -> `rom_paths`
- `AtariVectorEnv.game` -> `games`
- `AtariVectorEnv.get_action_set` -> `get_action_sets`

**Action space:**
`action_space` now covers all ROMs and no longer changes with batch size. Previously,
action space varied with batch size, breaking training scripts that held a reference to it.
`action_space` is now `MultiDiscrete`; `single_action_space` is `None` unless the full
action space is used, in which case it is always 16.

`TestVectorEnv::test_batch_size_async` is updated to reflect this.

### DIFF
--- a/src/ale/python/vector_env.py
+++ b/src/ale/python/vector_env.py
@@ -10,7 +10,7 @@ import numpy as np
 from ale_py import roms
 from ale_py.env import AtariEnv
 from gymnasium.core import ObsType
-from gymnasium.spaces import Box, Discrete
+from gymnasium.spaces import Box, Discrete, MultiDiscrete
 from gymnasium.vector import AutoresetMode, VectorEnv
 
 
@@ -19,9 +19,10 @@ class AtariVectorEnv(VectorEnv):
 
     def __init__(
         self,
-        game: str,
-        num_envs: int,
+        games: list[str] | str | None = None,
+        num_envs: int | None = None,
         *,
+        game: str | None = None,
         batch_size: int = 0,
         num_threads: int = 0,
         thread_affinity_offset: int = -1,
@@ -47,8 +48,9 @@ class AtariVectorEnv(VectorEnv):
         """Constructor for vector environment.
 
         Args:
-            game: ROM name
-            num_envs: Number of environments
+            games: List of ROM names
+            game: Single ROM name (used by gymnasium registration)
+            num_envs: Repeat each ROM this many times
             batch_size: If to provide a batch of environments (in async mode)
             num_threads: The number of threads to use for parallel environments
             thread_affinity_offset: The CPU core offset for thread affinity (-1 means no affinity, default: -1)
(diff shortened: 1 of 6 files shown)


AdrianOrenstein:
@pseudo-rnd-thoughts Ready for your review.

src/ale/python/vector_env.py
```diff
@@ -102,48 +110,60 @@ def __init__(
...
+            high=np.array([1.0, np.pi, 1.0]).astype(np.float32),
+            dtype=np.float32,
+            shape=(3,),
         )
-        self.action_space = gymnasium.vector.utils.batch_space(
-            self.single_action_space, self.batch_size
+        return single, gymnasium.vector.utils.batch_space(single, self.batch_size)
+
+    def _setup_discrete_action(self) -> tuple:
+        single = None if self.full_action_space is None else Discrete(16)
```
pseudo-rnd-thoughts: **Bug: `single_action_space` is always `Discrete(16)` — wrong value, and the `None` branch is dead.**

`self.full_action_space` is a `bool` (default `False`), so `self.full_action_space is None` is never true → `single_action_space` is unconditionally `Discrete(16)`. Two problems:
- The full action set has **18** actions (`PLAYER_A_*` 0–17 in `Constants.h`), not 16.
- For the common single/reduced-action case (`gym.make_vec("ALE/Breakout-v5", num_envs=10)`) this used to be `Discrete(4)`. Now `single_action_space.sample()` can return 4–15, which are invalid for Breakout, and it is inconsistent with `action_space` (`MultiDiscrete([4]*10)`). No test asserts `single_action_space`, so it slips through.

AdrianOrenstein: Thanks. I made these changes:
- When all of the action spaces are the same we set to the same action space. Added a test to VectorEnv.
- When different action spaces are seen we set single_action_space to None. Added a MultiRom test.

AdrianOrenstein:
@pseudo-rnd-thoughts please let me know if there's anything else to do for this PR 😄

jkterry1:
Hey just as a cursory question, could you help me understand what this is a feature that people would want? Perhaps I'm just missing something quite obvious.

AdrianOrenstein:
Hey @jkterry1, here are some reasons people would want multi-rom support to be part of the vector interface: 

1) Running many roms in parallel helps run more experiments at once. This feature increases the throughput by 40% compared to running multiple python processes. I have my experiment in [this blog post](https://adrianorenstein.com/blog/multirom/) where I ran 57 unique ROMs and multiple seeds per ROM.

image


2) If we can run multiple roms per environment instance, we can more easily evaluate agents that learn to play many games in their lifetime. Or agents that can switch between ROMs. I believe this PR will open up more challenging scenarios to evaluate RL algorithms.

jkterry1:
@AdrianOrenstein can you please get CI to pass?

AdrianOrenstein:
@jkterry1 Thanks for pinging me. It should be ready for review now.

jkterry1:
@AdrianOrenstein could you please rebase off main?

AdrianOrenstein:
@jkterry1 ready for CI

PR #707, opened 2026-06-28
Enable XLA on MacOS and fix XLA on GPUs

pseudo-rnd-thoughts:
This PR adds a test for XLA to check that on computers with GPUs the arrays are stored on GPUs and vice versa for CPUs (by default). 
Further, this PR enables XLA on MacOS, though CPU only (https://github.com/Farama-Foundation/Arcade-Learning-Environment/issues/704)
Plus, fixes Windows `pip install .` to use the custom triplet

I have tested this on Linux, macOS, and Windows (WSL and normal)

@szepi1991 this should fix your issues. If you can download a wheel from https://github.com/Farama-Foundation/Arcade-Learning-Environment/actions/runs/28333756257?pr=707 at the bottom and test that would be great


szepi1991:
Works across my Mac and Linux machines, thanks!

pseudo-rnd-thoughts:
Perfect, I'll merge and check if there are any other changes before cutting a release

PR #731, opened 2026-08-16
Reject non-positive frameskip in AtariVectorEnv

teddytennant:
`AtariEnv` rejects `frameskip <= 0` at construction. `AtariVectorEnv` never did, and the C++ skip loop is `for (skip_id = frame_skip_; skip_id > 0; --skip_id)`, so `frameskip=0` (and `-1`) constructs and then never advances the emulator.

    AtariVectorEnv("pong", num_envs=1, frameskip=0)  # used to succeed

Same check as `AtariEnv`: non-int type, or a non-positive int, raises `gymnasium.error.Error`.

`pytest tests/python/test_atari_vector_env.py::TestVectorEnv::test_invalid_frameskip` and `test_frameskip_warnings` on `AtariEnv` both pass.

Related to the same class of constructor check as #710.

### DIFF
--- a/src/ale/python/vector_env.py
+++ b/src/ale/python/vector_env.py
@@ -9,6 +9,7 @@ import gymnasium.vector.utils
 import numpy as np
 from ale_py import roms
 from ale_py.env import AtariEnv
+from gymnasium import error
 from gymnasium.core import ObsType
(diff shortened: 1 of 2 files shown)


pseudo-rnd-thoughts (approved):
Thanks for the PR, looks good

PR #733, opened 2026-08-23
Add doc pages for notable Atari environment sets

### LINKED ISSUE #708: Clearly Mark Different Sets of Atari Games In Documentation
The environments previously in Gym, the Atari 56 environments from DeepMind, and so on should be popped out in their own doc categorizations for researchers and students. Like some sort of page listed on the sidebar with links to the various notable particular sets of Atari environments

BolunDai0216:
## Description

Fixes #708

Adds sidebar-visible documentation pages listing the games in each commonly benchmarked subset of Atari environments, so multiple views exist simultaneously as suggested in the issue:

- **Atari 57** (`environments/atari_57`) — the 57 games of the DeepMind benchmark (DQN through Agent57), with references and a note that `Defender`/`Surround` were unavailable in `atari-py`, which is why some papers benchmark on only 55 of these games.
- **Previously in OpenAI Gym** (`environments/gym`) — the 62 games that were registered in OpenAI Gym, explaining the `Pong-v0` → `ALE/Pong-v5` naming migration and how the set relates to Atari 57 (all of it except `Surround`, plus `Adventure`, `AirRaid`, `Carnival`, `ElevatorAction`, `JourneyEscape` and `Pooyan`).
- **Complete list** — now included in a toctree (it previously wasn't in any), with cross-links to the set pages.

The Environments page gets a visible "Environment sets" section linking the three views, and the three pages live under a separate "Environment Sets" sidebar dropdown directly below Environments (per review, keeping the Environments dropdown to just the individual game pages).

The new pages follow the existing `complete_list` pattern: a checked-in `env-grid` HTML file embedded in a markdown page via `{raw} html`, using the same relative-link style (correct for the `dirhtml` builder used in CI). The grids were generated programmatically from the display names in `complete_list.html`, with assertions validating both set definitions and the relationship between them. Additional sets (e.g., the DQN Nature 49) can be added the same way if desired.

Sphinx build succeeds with no new warnings.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/environment-sets.md
+++ b/docs/environment-sets.md
@@ -1,0 +1,20 @@
+---
+firstpage:
+lastpage:
+---
+
+# Environment Sets
+
+Research papers rarely benchmark on all of the games the ALE supports, several well-known subsets are used instead.
+The pages below list the games belonging to each notable set:
+
+- [Complete list](environments/complete_list) — all 104 games supported by the ALE.
+- [Atari 57](environments/atari_57) — the 57 games used by DeepMind to benchmark agents, from DQN through Agent57.
+- [Previously in OpenAI Gym](environments/gym) — the 62 games that were registered in OpenAI Gym.
+
+```{toctree}
+:hidden:
+environments/complete_list
+environments/atari_57
+environments/gym
+```
--- a/docs/environments.md
+++ b/docs/environments.md
@@ -7,9 +7,6 @@ lastpage:
 
 ```{toctree}
 :hidden:
-environments/complete_list
-environments/atari_57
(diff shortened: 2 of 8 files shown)


pseudo-rnd-thoughts (changes requested):
Thanks @BolunDai0216 for the PR. Reviewing this locally, I don't like "Complete List", "Atari 57" and "Environments Previously in Gym" being in the table of contents
image

Could you remove these from the table of contents and rename "Environments Previously in Gym" to "Previously in OpenAI Gym"

